PR #29, opened 2023-07-30
Create school, school year, and cohort models

### LINKED ISSUE #13: Create school and cohort models
### School 
Id 
Name

### Cohort
Id
name: txt, null: false
school_id, null: false
school_year_id: int
grade: int, null: false
teacher_id: int, null: false
active: boolean, default: true

cohorts have a many to one relationship with schools

randallreedjr:
Resolves #13 

### Description

- Creates models for school, school_year, and cohorts, with accompanying migrations and factories
- Creates factory for user model
- Includes base specs that each new model has a valid factory

### Type of change

### Screenshots

N/A

### DIFF
--- a/app/models/school_year.rb
+++ b/app/models/school_year.rb
@@ -1,0 +1,2 @@
+class SchoolYear < ApplicationRecord
+end
--- a/app/models/school_year.rb
+++ b/app/models/school_year.rb
@@ -1,2 +1,3 @@
 class SchoolYear < ApplicationRecord
+  has_many :cohorts
 end
--- a/app/models/school_year.rb
+++ b/app/models/school_year.rb
@@ -1,3 +1,0 @@
-class SchoolYear < ApplicationRecord
-  has_many :cohorts
-end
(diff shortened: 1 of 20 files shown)


LizPrescott (approved):
Thanks for doing this!

app/models/school_year.rb
```diff
@@ -0,0 +1,3 @@
+class SchoolYear < ApplicationRecord
```
garettarrowood, line 1: Since the school year doesn't actually (and shouldn't) belong to a school, can we call it an AcademicYear instead?

db/migrate/20230730151848_create_cohorts.rb
```diff
@@ -0,0 +1,14 @@
+class CreateCohorts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :cohorts do |t|
+      t.text :name, null: false
+      t.references :school, null: false, foreign_key: true
+      t.references :school_year, null: false, foreign_key: true
+      t.integer :grade, null: false
```
garettarrowood, line 7: Just to not confuse this with grades of A,B,C,D,F, etc, could we call this field "grade_level"?

garettarrowood (changes requested):
See comments above

garettarrowood:
@randallreedjr - Do you think you'll be able to get back to this this week? And if not, are you cool with me just pulling what you've done so far and applying the last changes suggested above?

randallreedjr:
> @randallreedjr - Do you think you'll be able to get back to this this week? And if not, are you cool with me just pulling what you've done so far and applying the last changes suggested above?

@garettarrowood Done!

PR #30, opened 2023-07-30
Fix the entrypoint to actually run migrations

cflipse:
Quotes around the comparison string mean that the `*` is not globbed


jasonodoom:
LGTM.

jasonodoom (approved):

PR #32, opened 2023-07-30
Stock API Integration

marc:
Add marketstack api to fetch stock prices.

Resolves need for api integration. (Issue does not exist.)

### Description

Integrate with MarketStack API to fetch stock prices.
See https://marketstack.com/documentation#eod_data for ore information.
Using the free tier so we only have access to the end of day data.

### Type of change

### Screenshots

n/a


LizPrescott:
I dig the data model!

app/models/stocks/marketstack_api.rb
```diff
@@ -0,0 +1,74 @@
...
+
+# Get stock prices from api with:
+# api = Stocks::MarketstackApi.new(api_key: Rails.application.credentials.marketstack_key)
+# api.get_end_of_day_price("KO,SNE,TWX,DIS,SIRI,F,EA,FB,UA,LUV,GPS")
+
+class Stocks::MarketstackApi
+  attr_accessor :api_key
+
+  def initialize(attrs)
+    @api_key = attrs[:api_key] if attrs[:api_key].present?
```
cflipse, line 12: can we make this a keyword arguement?

```ruby
def initialize(api_key: nil)
  @api_key = api_key
```

app/models/stocks/marketstack_api.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def get_end_of_day_price(symbols, date = "latest")
```
cflipse, line 15: likewise, for date, rather than a positional argument

and some kind of documentation on what's valid for a `date: ` param

app/models/stocks/marketstack_api.rb
```diff
@@ -0,0 +1,74 @@
...
+    return if symbols.blank?
+    return if @api_key.blank?
+    uri_params = {
+      access_key: @api_key,
+      symbols: symbols,
+      date: date,
+      limit: symbols.count(",") + 1
+    }
+    uri = URI("http://api.marketstack.com/v1/eod")
```
cflipse, line 24: maybe move the URI up to a const, or make it injectable (with default) at the initializer?

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize(attrs = {})
+    assign_attrs(attrs) unless attrs.blank?
+  end
+
+  def get_end_of_day_price(symbol:, date: "latest")
+    # date could be "latest" or "YYYY-MM-DD"
+    redis = Redis.new(url: REDIS_HOST)
+    cached = redis.get("eod_#{symbol.downcase}_#{date}")
+    assign_attrs(JSON.parse(cached)) if cached.present?
```
cflipse, line 25: let's keep all of the transport details in the api model, rather than letting the serialized JSON leak out for another thing to have to deal with.

cflipse: Also, Rails can handle the full cache assignment logic:
https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching

```
Rails.cache.fetch( .... ) do
  fetch the json
end
```

app/models/stocks/marketstack_api.rb
```diff
@@ -0,0 +1,74 @@
...
+    uri.query = URI.encode_www_form(uri_params)
+    json = Net::HTTP.get(uri)
```
cflipse, line 26: If you wanted to add Faraday, which presents a much less irritating HTTP api, that would be :+1:

app/models/stocks/marketstack_api.rb
```diff
@@ -0,0 +1,74 @@
...
+    parsed.each { |stock_price| stock_price.write_to_cache(is_latest: true) }
+  end
+
+  def parse_eod_response(prices)
+    parsed = []
+    prices.each do |api_price|
+      price = Stocks::Price.new(api_price)
+      parsed << price
+    end
+    parsed
```
cflipse, line 38:
```ruby
prices.map { |price| Stocks::Price.new(price) }
```

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
+require "json"
+require "redis"
+
+class Stocks::Price
+  REDIS_HOST = "redis://redis:6379/"
```
cflipse, line 5: We should use the configure this in the application config,  enabling the redis cache, and use that.

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
...
+    attrs.each_pair do |k, v|
+      instance_variable_set("@#{k}", v) if v.present? && respond_to?(k)
+    end
+  end
+
+  def initialize(attrs = {})
+    assign_attrs(attrs) unless attrs.blank?
+  end
+
+  def get_end_of_day_price(symbol:, date: "latest")
```
cflipse, line 21: fetching a different price, from an _instance_ of a price, feels like an unintuitive approach.  This might be better as a class method, or located on the API model.

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def assign_attrs(attrs = {})
+    attrs.each_pair do |k, v|
+      instance_variable_set("@#{k}", v) if v.present? && respond_to?(k)
+    end
+  end
+
+  def initialize(attrs = {})
+    assign_attrs(attrs) unless attrs.blank?
+  end
```
cflipse, line 19: This can all be processed by making this an `ActiveModel` https://guides.rubyonrails.org/active_model_basics.html#model

cflipse:
Bunch of comments for cleanup, but good to merge.

garettarrowood (approved):
As Chris mentioned, let's get this out and loop back for the upgrades. I'll open a PR to make these edits myself.

PR #37, opened 2023-07-30
Stock prices integration

marc:
Pull stock prices from redis cache

### Type of change

### Screenshots

image


cflipse:
Once #40 is merged, the CI failure should be addressed if you bring it up to date with main.  The issue there was that CI was running without a redis, while this is trying to directly connect to a running instance.

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize(attrs = {})
+    assign_attrs(attrs) unless attrs.blank?
+  end
+
+  def get_end_of_day_price(symbol:, date: "latest")
+    # date could be "latest" or "YYYY-MM-DD"
+    redis = Redis.new(url: REDIS_HOST)
+    cached = redis.get("eod_#{symbol.downcase}_#{date}")
+    assign_attrs(JSON.parse(cached)) if cached.present?
```
cflipse, line 25: Please use the Rails cache wrappers rather than directly connecting to a running redis instance.

see the guides here:  https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

app/models/stocks/price.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def assign_attrs(attrs = {})
+    attrs.each_pair do |k, v|
+      instance_variable_set("@#{k}", v) if v.present? && respond_to?(k)
+    end
+  end
+
+  def initialize(attrs = {})
+    assign_attrs(attrs) unless attrs.blank?
+  end
```
cflipse, line 19: These can be replaced with `ActiveModel::Model` , which should give us another handful of benefits as well:

https://guides.rubyonrails.org/active_model_basics.html#model

cflipse (changes requested):
Please use the cache wrappers rather than directly connecting to redis.  They're built for this sort of thing, and it keeps the overall application configuration from leaking out into random source files.

garettarrowood:
Closing for now.

The MarketStack integration has been partially merged. A writeup of the current state of things can be found at #56 . This will be a helpful reference to reach the final StockTicker state.

PR #39, opened 2023-08-01
Attendance model

### LINKED ISSUE #17: Create attendance table / model
**Attendance**
user_id INT, null: false
school_week_id INT, null: false
school_period_id, INT
verified BOOL, default: false, null: false
attended BOOL, default: false, null: false
quarter_bonus INT, limit: 1

**school_weeks** (table)
school_period_id, INT
week_number INT default: 0
cohort_id INT
week_id INT

Perfect attendance will be used to give students bonuses to buy stocks

kristenking:
Resolves #17 

## Description
This work-in-progress PR includes Attendance and SchoolWeek models. The Attendance model has associations with the User and SchoolWeek models. As the SchoolWeek model was not yet present in the project, I had to create it as part of implementing the Attendance model. 

Given the dependencies between Attendance and SchoolWeek, we might encounter potential issues that need to be addressed. I am seeking early feedback on these model definitions before I write the associated spec tests. 

## Type of change
This PR adds new functionality (non-breaking change).


garettarrowood:
@kristenking - This looks good so far! I'd have to see our central doc to remind myself of all of our other considerations at this time.

Any ideas what `quarter_bonus` is supposed to represent? I see you haven't stored it as a boolean, but it looks like that is all it is supposed to represent. If so, I'd suggest we just name it `bonus_awarded` and change it over to the boolean.

cflipse:
> stored it as a boolean, but it looks like that is all it is supposed to represent. If so, I'd suggest we just name it bonus_awarded and change it over to the boolean.

Also, if it's just an indiciator if the bonus has been awarded, we can make it a little bit richer by turning it into a timestamp:  `bonus_awarded_at`  -- if it's null, then the bonus hasn't been given, and if it's populated, it indicates not only that a bonus _has_ been awarded, but also _when_.

cflipse:
(also, since the PR message mentions this being a work in progress, I converted it to a draft)

app/models/attendance.rb
```diff
@@ -0,0 +1,11 @@
+class Attendance < ApplicationRecord
+  belongs_to :user
+  belongs_to :school_week
+
+  validates :user, presence: true
+  validates :school_week, presence: true
+  validates :school_period_id, presence: true, numericality: {only_integer: true}
```
cflipse, line 7: This reads like it's a foreign key.  Do we have a `SchoolPeriod` class, or does this represent something else?

app/models/attendance.rb
```diff
@@ -0,0 +1,11 @@
...
+  validates :verified, inclusion: {in: [true, false]}
+  validates :attended, inclusion: {in: [true, false]}
+  validates :quarter_bonus, numericality: {only_integer: true, greater_than_or_equal_to: 0, less_than_or_equal_to: 1}
```
cflipse, line 10: (discussed in main thread, but this will look to rails exactly like it's a boolean, and so is probably unecessary as well)

app/models/attendance.rb
garettarrowood: Looks like there are `SchoolPeriod`s in the original application, but I'm not sure what they represent yet.

app/models/school_week.rb
```diff
@@ -0,0 +1,3 @@
+class SchoolWeek < ApplicationRecord
+    has_many :attendances
```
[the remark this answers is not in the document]
garettarrowood: I'm looking at the original schema now... and it's quite a mess. SchoolWeeks look like
```
    t.integer "school_period_id"
    t.integer "week_number", default: 0
    t.integer "week_in_period"
    t.integer "class_id", null: false
    t.integer "week_id"
  ```
I believe we need `SchoolPeriod`s, which embodies a quarter and is labeled such in the UI. Then...

There is a `school` (that belongs to an `academic_year`)
`school_period`s belong to a school.
`school_week`s belong to a `school_period`, and then a school through `school_period`
`cohorts` belong to a `school_period` and `school` (although they could just belong to a `school` through `school_period`).

It makes sense that an attendance (which I'd be open to renaming to `attendance_week`, `user_attendance`, `student_attendance` or anything a little more semantic) belongs to a user and a school week. I'm not sure of the benefit of duplicating the `school_period` association on it. 

Looks like the old code iterates through all the weeks in the period/quarter, then checks if `attendances.verified` and `attendances.attended` are true, then updates ALL of the weeks again with `quarter_bonus = 1`. What a mess.

May make the most sense to have `week_attendances` and `period_attendances` then store the `qualified_for_bonus_at` on `period_attendances`. Another possibility would be to just store perfect attendance bonus on the report card class, which should embody a period/quarter and point at a student already.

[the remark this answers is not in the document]
Comment (garettarrowood) [reply app/models/school_week.rb:2]: After talking this out with @suttondemlong @naomiyocum & @kristenking , let's punt `quarter_bonus` altogether.

File: db/migrate/20230730155811_create_attendances.rb [hunk 4]
```diff
@@ -0,0 +1,14 @@
+class CreateAttendances < ActiveRecord::Migration[7.0]
+  def change
+    create_table :attendances do |t|
+      t.references :user, null: false, foreign_key: true
+      t.references :school_week, null: false, foreign_key: true
+      t.integer :school_period_id
+      t.boolean :verified, default: false, null: false
+      t.boolean :attended, default: false, null: false
+      t.integer :quarter_bonus, limit: 1
```
garettarrowood, line 9: So this field in particular.

garettarrowood:
Let's create the `weeks` table and :

```
  create_table "weeks", force: :cascade do |t|
    t.integer "academic_year_id", null: false
    t.date "start_date", null: false
  end
```

and the `school_periods`:
```
  create_table "school_periods", force: :cascade do |t|
    t.integer "period_number"
    t.integer "cohort_id", null: false
  end
```

app/models/school_week.rb
```diff
@@ -0,0 +1,7 @@
+class SchoolWeek < ApplicationRecord
+  belongs_to :school_period
+
+  has_many :attendances
```
garettarrowood, line 4:
```suggestion
  has_many :student_attendances
```

db/migrate/20230730145516_create_schools.rb
```diff
@@ -2,6 +2,7 @@ class CreateSchools < ActiveRecord::Migration[7.0]
   def change
     create_table :schools do |t|
       t.text :name, null: false
+      t.integer :academic_year_id, null: false
```
garettarrowood, line 5: We can't alter a previously run migration, because it's already been played. Once migration files are merged to the main branch of a repository, and then (assumed) run against the production database, they become dead code files. The only way to add the `academic_year_id` to `School` is to create a new migration that only adds a column.

app/models/student_attendance.rb
```diff
@@ -0,0 +1,7 @@
+class StudentAttendance < ApplicationRecord
+  belongs_to :user
```
garettarrowood, line 2: Since we have to go back and fix the migration commented on above, let's make one more tweak. Let's rename the `user_id` column on `student_attendance` to `student_id`.

This association would then be defined like this:
```ruby
  belongs_to :student, class_name: "User", foreign_key: :student_id
```

garettarrowood:
Thank you for sorting through these associations and getting these models together!

PR #41, opened 2023-08-02
Add role to user

### LINKED ISSUE #28: Add role to User model
--------

# Summary

Users need to have a "role" that can be used for permissions. These roles will be: admin, teacher, student, alumni. Users may have more than one role. 

# Criteria for Completion

 - [x] role has been added to user model
 - [x] role is an enum array
 - [x] role cannot be anything but ADMIN, TEACHER, STUDENT, ALUMNI
 - [x] appropriate tests have been written

-------

naomiyocum:
Resolves #28 

### Description

- Include summary of change and what issue is fixed, including relevant motivation and context.
  - Role enum added to User model (admin, alumni, student, teacher)
  - Added user factory to assist with testing
- Explain the need for any added dependencies for this change.
  - Added shoulda-matchers gem to assist with testing
  - Added pry gem
- How have you tested this?
  - `define_enum_for` matcher - last test in user_spec.rb
- Anything else we should know about?
  - Unsure how a user will have more than one role (as described in the issue)

### Type of change

### Screenshots

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,9 +77,11 @@ end
 group :development, :test do
   gem "dotenv-rails"
   gem "brakeman"
+  gem "shoulda-matchers"
   gem "standard", require: false
   gem "standard-rails", require: false
   # RSpec behavioral testing framework for Rails.
+  gem "pry"
   gem "rspec-rails", "~> 6.0.3"
   gem "factory_bot_rails"
   gem "faker"
--- a/Gemfile
+++ b/Gemfile
@@ -81,7 +81,7 @@ group :development, :test do
   gem "standard", require: false
   gem "standard-rails", require: false
   # RSpec behavioral testing framework for Rails.
-  gem "pry"
+  gem "pry-byebug"
   gem "rspec-rails", "~> 6.0.3"
   gem "factory_bot_rails"
   gem "faker"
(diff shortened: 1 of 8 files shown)


garettarrowood:
This look great. Nice work!

I've not done this myself before, but I believe we wanted to set up an array column so that users can have multiple roles in the future (teacher/admin being the most likely). [Here's an article I just found about that](https://dev.to/kputra/rails-postgresql-array-1jn0), but perhaps you can find something better.

Good call adding `pry`. Could we add `pry-byebug` instead though? That's a wrapper library around incuding both byebug and pry in the repo. Just gives devs more options if their default debugger is byebug.

naomiyocum:
I will add `pry-byebug` instead!

I'm doing research on creating an array for the role and I believe if we want to have an array of values, we cannot use enum? It seems enum is limited to one value. I have found some out of date gems that solved this problem - [bitwise](https://medium.com/jobteaser-dev-team/rails-bitwise-enum-with-super-powers-5030bda5dbab) and [array_enum](https://github.com/freeletics/array_enum).

garettarrowood:
Since we don't have a clear path forward with an array enum, and we don't know for sure that it would ever be needed, let's just move forward with a standard enum.

garettarrowood (approved):

Issue #7, opened 2023-07-29
Enable and configure dependabot


cflipse:

jasonodoom:
I will commit the `dependabot.yml` but either @cflipse , @LizPrescott , @garettarrowood  or whomever else with admin privileges for this project should enable the below option(s).

![](https://github.blog/wp-content/uploads/2020/05/enable-automated-security-updates.gif?resize=1024%2C619)

garettarrowood:
Looks like Github has moved around all of its settings. Here's what the dependabot settings look like. Perhaps it's on by default now? Unsure.

<img width="1147" alt="Screenshot 2023-08-01 at 8 33 23 AM" src="https://github.com/rubyforgood/stocks-in-the-future/assets/12770108/5ffe41dd-16b1-407d-b14f-56fa7eca2334">

garettarrowood:
Actually - it's just in another spot. -- I've enabled it:

<img width="1227" alt="Screenshot 2023-08-01 at 8 36 38 AM" src="https://github.com/rubyforgood/stocks-in-the-future/assets/12770108/6ca8916c-820c-4d68-a2d3-229ec9f33110">

When I go to turn on it opening PRs on a weekly cadence, it prompts me with the ability to add a commit with the dependabot.yml file with this in it:

```
# To get started with Dependabot version updates, you'll need to specify which
# package ecosystems to update and where the package manifests are located.
# Please see the documentation for all configuration options:
# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

version: 2
updates:
  - package-ecosystem: "" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "weekly"
```

Adding that file may be what turns it on.

jasonodoom:
> Actually - it's just in another spot.
> <img alt="Screenshot 2023-08-01 at 8 36 38 AM" width="1227" src="https://user-images.githubusercontent.com/12770108/257515636-6ca8916c-820c-4d68-a2d3-229ec9f33110.png">
> 
> When I go to "enable" opening PRs on a weekly cadence for version updates, it prompts me with the ability to add a commit with the dependabot.yml file with this in it:
> 
> ```
> # To get started with Dependabot version updates, you'll need to specify which
> # package ecosystems to update and where the package manifests are located.
> # Please see the documentation for all configuration options:
> # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
> 
[...]

PR #43, opened 2023-08-02
Add necessary file to enable dependabot in Settings

garettarrowood:
To enable Dependabot, this yml file must first exist. This file is not yet configured. @jasonodoom has interest in setting this up properly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


garettarrowood:
This does not resolve #7 , but just unblocks it.

jasonodoom:
@cflipse can you please provide input so I will know how to proceed?

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
```
cflipse, line 8: this should specify as `bundler`

cflipse:
Sure, go for it.  

I don't _think_ we're using anything other than the bundler ecosystem, though _maybe_ the docker ecosystem is worth adding?  I'm not terribly familiar with what this will do with just `docker-compose` files though, so that might be pointless.

.github/dependabot.yml
jasonodoom: glad I waited for your feedback :)

PR #82, opened 2023-10-27
Rails 7.1

cflipse:
### Description

Update application from Rails 7.0 to Rails 7.1

Note that there's a deprecation warning on `application.secrets` that comes from within Devise; it's not an issue at the moment, but at some point before moving to 7.2, devise would need to be updated.

### Type of change

* Housekeeping


garettarrowood (approved):
Welcome back! LGTM

PR #123, opened 2024-03-07
Adding User, Stock, and Portfolio models

ramirez-gabriela27:
This PR adds a User, Portfolio, and Stock model, as well as a view for the portfolio.


app/models/user.rb
```diff
@@ -0,0 +1,3 @@
+class User < ApplicationRecord
```
maks112v, line 1: There will probably be a merge conflict with the devise user

app/views/layouts/application.html.erb
```diff
@@ -6,7 +6,7 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%# stylesheet_link_tag "application", "data-turbo-track": "reload" %>
```
maks112v, line 9: Why is this getting commented out?

ramirez-gabriela27: I commented it out because it was causing an error when running `rails server`, just forgot to put it back!

h-m-m:
It looks like a lot of this work was merged in with a different PR

PR #116, opened 2024-02-06
Reinit Project

maks112v:
Resolves #000 

### Description

- Include summary of change and what issue is fixed, including relevant motivation and context.
- Explain the need for any added dependencies for this change.
- How have you tested this?
- Anything else we should know about?

### Type of change

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)
* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* Documentation update

PR #132, opened 2024-05-21
Pull in earlier config files

h-m-m:
### Description

This pulls a few select config files files prior to https://github.com/rubyforgood/stocks-in-the-future/pull/116

### Details

I'm opening this initially as a draft to see what, if any, of the github workflow features will work with out much effort.

I also think we should spend at least a minute or two considering if any of these are no longer appropriate


.ruby-version
```diff
@@ -1 +1 @@
-ruby-3.2.3
+ruby-3.2.2
```
maks112v, line 1: Why is the ruby version being downgraded, are there issues with the higher version?

maks112v (approved):

.ruby-version
h-m-m: I put it in the commit message, and I'll copy it here, too.

It looks like `ubuntu-latest` in GH actions is the latest patch of LTS
22, not LTS 24. I probably could add something to get updated packages
so 3.2.3 is available. I want to first check if the project has any
other problems before getting into the weeds

h-m-m: This is part of the reason I have this PR still marked as draft

PR #136, opened 2024-05-31
Developer tooling updates!

abachman:
Goal: make sure `bin/dev` and `bin/setup` work.

This PR does three things: 

1. switch to a default Rails setup where there is no local config/database.yml, only a config/database.yml.sample
2. ignore the local .vscode folder
3. replace `bun` with `yarn`

Justifications:

1. some systems require different database.yml configuration, so it can't be in a committed file
2. i use different vscode color themes for each project i open, so vscode keeps creating create settings files locally
3. `yarn` is more common for Rails projects and lets us use Rails' native esbuild support through the `jsbundling-rails` gem.


config/database.yml.sample
```diff
@@ -15,6 +15,8 @@
 default: &default
   adapter: postgresql
   encoding: unicode
+  # running postgres on docker requires specifying an explicit host
+  # host: 127.0.0.1
```
h-m-m, line 19: True. Also you can't specify the host as `localhost` or it will switch to sockets instead of TCP

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "app",
   "private": true,
   "scripts": {
-    "build": "bun bun.config.js",
+    "build": "esbuild app/javascript/*.* --bundle --sourcemap --format=esm --outdir=app/assets/builds --public-path=/assets",
```
h-m-m, line 5: 🎉

h-m-m (approved):
LGTM

Issue #146, opened 2024-05-31
Add function that will pull API and Implement API changes at EOD.


JAYKAY2291:
# Summary

Function needed that will pull API at the end of each day and implement price changes.

## Things out of Scope
This function is independent of adding buy orders which is called out in issue #145, although this function will be working in tandem with the buy order function. 

# Criteria for Completion
> This should be a bulleted list of items that, if satisfied, would result in an acceptably complete Pull Request. Be as specific as you need to be. For example:

 - [x] At the end of each day, API should be pulled, and displayed stock prices should be changed. 
 - [x] This feature must be available for all users to view.
 - [x] Add a test proving that it works

abachman:
related issue in #56

PR #151, opened 2024-05-31
#146 Api demo data

LizPrescott:
This is just the beginnings of addressing issue #146, including:
 the demo data from https://www.alphavantage.co/documentation/ as it will come back from 
https://www.alphavantage.co/query?function=GLOBAL_QUOTE&symbol={symbol}&apikey={api_key}
The scaffolding for a job that would call this endpoint for each symbol in a json or cvs file where we'll store the small group of stock symbols we need to update.


abachman (approved):
Dig it, https://www.alphavantage.co/ seems decent.

PR #155, opened 2024-05-31
Basic developer tooling for running the app in Docker

abachman:
A proposed docker-compose.yml file for running and developing the app in docker and some supporting documentation and simple tooling to use it. 

### install yarn with `npm install -g yarn`

In `Dockerfile.dev`, use `npm install -g yarn` to get yarn instead of Yarn's `corepack` recommendation. This lets us not have to reinstall corepack and yarn every time we run a command like `docker compose run stocks rails test`.

```dockerfile
# Install packages needed to build gems
RUN apt-get update -qq && \
    apt-get install --no-install-recommends -y \
    build-essential libpq-dev curl \
    nodejs npm
RUN npm install -g yarn

COPY Gemfile Gemfile.lock ./
RUN bundle install

COPY package.json yarn.lock ./
RUN yarn
```

### use `node:lts-alpine` as the image for stocks-js and stocks-css services

It might be possible to reuse the `stocks` service image to build assets, but `node:lts-alpine` has yarn preinstalled, so there's very little fiddling required to make the services work with `docker compose up`.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -12,13 +12,13 @@ RUN apt-get update -qq && \
     apt-get install --no-install-recommends -y \
     build-essential libpq-dev curl \
     nodejs npm
+RUN npm install -g yarn
 
 COPY Gemfile Gemfile.lock ./
 RUN bundle install
 
 COPY package.json yarn.lock ./
-RUN npm install -g yarn
-RUN yarn install
+RUN yarn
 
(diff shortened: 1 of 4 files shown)


BombDefusingSloth (approved):
Works for me on Windows 10.

spolisar (commented):
Works for me on Linux.

PR #161, opened 2024-06-01
Ruby test and lint github action workflow

abachman:
This PR introduces a workflow for running the Rails test suite and [Ruby Standard](https://github.com/standardrb/standard) linter on pull requests for the project.

This workflow _works_ as currently written, but fails checks because we have test failures and linter violations.

Once those issues are fixed, we can add it as a branch protection in the project settings.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       POSTGRES_PASSWORD: password
       POSTGRES_DB: sif
     ports:
-      - 5432
+      - 5432:5432
     healthcheck:
       test: ["CMD-SHELL", "pg_isready", "--database", "stocks_in_the_future_development" ]
       interval: 10s
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,64 @@
+---
+name: "Continuous Integration"
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
       POSTGRES_PASSWORD: password
       POSTGRES_DB: sif
     ports:
-      - 5432
+      - 5432:5432
```
jonny5, line 22: is this intended? (not too familiar with myself)

jonny5 (approved):
One small comment, LGTM if that is intended 👍

docker-compose.yml
h-m-m: I think one port is the port inside the container and the other is outside the container (or the other way around)

abachman: Yes! This lets the postgres docker image be useable from your host machine. 

So if you're comfortable with docker, you could run `docker compose up -d db` from this project directory and have Postgres without having to install it on your OS directly.

abachman: in docker configuration--ports and volumes in particular--the colon-separated `thing:thing` refers to `outside:inside` or `host:image`. 

In this case, `5432:5432` read left to right says, "expose on port 5432 on my host machine the port 5432 from inside the docker container". 

When we use a single value like `5432`, it means "the docker container I'm describing should expose port 5432 to other containers on the same network." "The same network" here means the other services named in docker-compose.yml.

PR #165, opened 2024-06-01
Setup delayed_job and basic configuration

americac:
Gems installed:
- daemons
- delayed_job_active_record

Updated Procfile.dev to include running delayed_job


abachman (approved):
:+1: 

Gemfile changes line up with standardrb, where we're headed anyways

PR #166, opened 2024-06-01
Job now prints full output for all api calls. Next step, extract just…

LizPrescott:
… the ticket symbol and closing price

### DIFF
--- a/app/jobs/stock_prices_update_job.rb
+++ b/app/jobs/stock_prices_update_job.rb
@@ -4,5 +4,18 @@ class StockPricesUpdateJob < ApplicationJob
   def perform(*args)
     # For each stock symbol, request the latest closing cost
     # update the stocks table with each new closing cost
+    stock_symbols = ["KO", "SNE", "TWX", "DIS", "SIRI", "F", "EA", "FB", "UA", "LUV", "GPS"]
+    stock_symbols.each do |symbol|
+      api_request(symbol)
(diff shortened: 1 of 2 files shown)


app/jobs/stock_prices_update_job.rb
```diff
@@ -4,5 +4,18 @@ class StockPricesUpdateJob < ApplicationJob
   def perform(*args)
     # For each stock symbol, request the latest closing cost
     # update the stocks table with each new closing cost
+    stock_symbols = ["KO", "SNE", "TWX", "DIS", "SIRI", "F", "EA", "FB", "UA", "LUV", "GPS"]
+    stock_symbols.each do |symbol|
+      api_request(symbol)
+      stock_db = Stock.find_by(ticker: symbol)
+    end
```
h-m-m, line 11: Maybe ensure the stock can be found in the DB before we request the stock? As written, would this cause the job to abort if the stock with one of the tickers can't be found, skipping the processing of any further stocks in the list

abachman: `.find_by` would be okay here, since it returns `nil` instead of raising NotFoundError

abachman (approved):

app/jobs/stock_prices_update_job.rb
h-m-m: Oh, I forgot about that, thanks.

This means we'll have to do nil checks, though

PR #168, opened 2024-06-01
Fix all standardrb errors and warnings

abachman:
In order to establish a safe base for auto-formatting using editor-based tooling with the Standard formatter, I want to clean up all existing `standardrb` and `standardrb-rails` errors and warnings.

I did this in two stages: 
1. `bundle exec standardrb --fix`
2. Manually fix everything that autoformat couldn't touch. This was mostly Rails problems.

This way we can add "Ruby Lint" as a required check for all Pull Requests going forward.


README.md
```diff
@@ -40,6 +42,38 @@ Access the app via `localhost:5000`
...
+
+We use Minitest for testing. You can run the test suite like this
+
+```console
+$ bin/rails db:create    # one time setup
+$ bin/rails test
+
+# in docker
+$ docker compose run stocks bin/rails db:create
+$ docker compose run stocks bin/rails test
```
h-m-m, line 64: Good, useful

app/controllers/schools_controller.rb
```diff
@@ -8,23 +8,25 @@ def index
   end
 
   # GET /schools/1 or /schools/1.json
-  def show; end
+  def show
+  end
```
h-m-m, line 12: I like the single lines aesthetically, but I'm not going to fight the standardrb defaults over it

app/models/classroom.rb
```diff
@@ -1,5 +1,5 @@
 class Classroom < ApplicationRecord
   belongs_to :year
   belongs_to :school
-  has_many :users
+  has_many :users, dependent: :nullify
```
h-m-m, line 4: I thought through the alternative arguments to `dependent:` and I think this is the correct one

app/models/stock.rb
```diff
@@ -1,6 +1,6 @@
 class Stock < ApplicationRecord
-  has_many :portfolio_stocks
-  has_many :orders
+  has_many :portfolio_stocks, dependent: :nullify
+  has_many :orders, dependent: :nullify
```
h-m-m, line 3: This feels wrong to me because stocks should never be deleted. They should be moved to an `inactive` state instead. Maybe `:restrict_with_error` instead?

app/models/portfolio.rb
```diff
@@ -1,7 +1,7 @@
 class Portfolio < ApplicationRecord
   belongs_to :user
-  has_many :portfolio_transactions
-  has_many :portfolio_stocks
+  has_many :portfolio_transactions, dependent: :destroy
+  has_many :portfolio_stocks, dependent: :destroy
```
h-m-m, line 4: This may also be a candidate for `restrict_with_error` or `restrict_with_exception`

h-m-m:
I like autoformatters because they help me have a second chance to look at the code with a fresh perspective in addition to their other benefits. This looks good to me. I especially like all the README updates

app/controllers/schools_controller.rb
abachman: :D  

Now you're feeling that ["unconfigurable configuration"](https://github.com/standardrb/standard#:~:text=and%20provides%20an-,unconfigurable%20configuration,-to%20all%20of) vibe.

app/models/stock.rb
abachman: I like it. 

"should never be deleted" will have a lot more interesting considerations in the future, but this is an easy fix now

app/models/classroom.rb
abachman: I also suspect we may end up with a join model here.

abachman:
This is something we can pick back up after the event. Closing for now, can reopen a new PR later.

app/models/classroom.rb
h-m-m: 🤔 Yeah, I can see that

Issue #163, opened 2024-06-01
Permissions creation


h-m-m:
# Summary

Create a set of basic permissions we'll know we need

## Things out of Scope

We don't need to put every permission we can think of in this ticket. Feel free to break chunks of permission settings out into other issues that link back to here

## Things to Consider

We have a consensus that the Pundit gem is suitable

# Criteria for Completion

 - [ ] Students can place orders for their portfolio and not for any other students

h-m-m:
This requires #162 to exist first. This can be a good prerequisite step before doing #153

pollygee:
@jonny5 we are getting this project moving again and could use as much help as possible.  Are you still interested in contributing?

jonny5:
@pollygee Glad that there is work being done on the project! Unfortunately I don't have a ton of bandwidth these days, may have a little time later in the summer but can't promise anything

PR #169, opened 2024-06-01
Basic pundit setup with OrderPolicy on create

jonny5:
- Needs to be integrated into a (currently not existing) controller with the line `authorize @order` to be useful

For issue: https://github.com/rubyforgood/stocks-in-the-future/issues/163

### DIFF
--- a/app/policies/order_policy.rb
+++ b/app/policies/order_policy.rb
@@ -1,0 +1,5 @@
+class OrderPolicy < ApplicationPolicy
+  def create?
+    user.portfolio == resource.portfolio
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -41,3 +41,5 @@ group :test do
   gem "selenium-webdriver"
   gem "webmock"
 end
+
+gem "pundit", "~> 2.3"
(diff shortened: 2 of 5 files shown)


app/policies/order_policy.rb
```diff
@@ -0,0 +1,12 @@
+class OrderPolicy
+  attr_reader :user, :order
+
+  def initialize(user, order)
+    @user = user
+    @order = order
+  end
```
h-m-m, line 7: If you want, you can create a `BasePolicy` class that other classes inherit from so we don't have to redefine `initialize` every time, e.g.:

https://github.com/18F/identity-dashboard/blob/8b06d08356cd0bf21558613168001987c5d4e0a4/app/policies/base_policy.rb#L1-L7

abachman (approved):

app/policies/order_policy.rb
jonny5: 👍  wasn't sure what defaults would be good, I ran `rails g pundit:install` which sets up the basic defaults and am now inheriting from that policy

PR #170, opened 2024-06-01
Ignore the Ruby Lint job in CI for now

abachman:
We don't need this yet, and there are a lot of folks whose work would be negatively impacted if they had to merge a bunch of conflicts. 

We can pick it back up after the event. 🙌🏻

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,22 +8,27 @@ on:
     types: ['opened', 'reopened', 'synchronize', 'unlocked']
 
 jobs:
-  ruby-lint:
-    name: Ruby Lint
-    runs-on: ubuntu-latest
+  # NOTE: @abachman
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,22 +8,27 @@ on:
...
 
 jobs:
-  ruby-lint:
-    name: Ruby Lint
-    runs-on: ubuntu-latest
+  # NOTE: @abachman
+  #   In order to move a little faster for the Ruby for Good weekend event, I'm
+  #   commenting out the ruby-lint job. We can re-enable it after the event
+  #   once we've had a chance to let the dust settle and run `standardrb --fix`
+  #   on everything.
```
h-m-m, line 15: Thank you for adding this note

h-m-m (approved):

Issue #145, opened 2024-05-31
Create function that implements Daily task of executing trades


JAYKAY2291:
# Summary

Add function that executes trades made by student accounts. The trades will be seen as "pending" by student user and not "completed" until the end of the day when all trades will be finalized.

## Things out of Scope
Adding API is not part of this issue.

## Things to Consider
Stock is bought at the price at the the time of trade and API change should not be taken into account. 

# Criteria for Completion
 - [ ] When a Student purchases stock, it will initiate a trade.
 - [ ] Trade will be listed as "pending" until end of day. 
 - [ ] At end of day, "pending" trade will convert to "completed trade."
 - [ ] This feature should only be available to student users.
 - [ ] Add an automated test proving that it works

pollygee:
@americac @jonny5 are either of you interested in still helping out with this issue?

DevelopMan:
@pollygee The scope is not clear for this issue.
Do I need to create a task to collect all pending orders for each student and execute them at 23:00 (server time) using the order price?
I'd be appreciated to have some documentation that explains "execution" process. As far as I understand there should be transaction and portfolio updates, but I can miss something.

pollygee:
@DevelopMan sorry about that you're right there isn't enough info in a lot of these tickets.  I'm closing them and will open new tickets.

PR #174, opened 2024-06-01
Execute trades job issue 145

jonny5:
Issue #145 

- Creates a service to process pending orders and wrapper job to run that service for all pending orders. 
- Changed floats to decimal for consistency
- Linked order to respective `portfolio_transaction` and `portfolio_stock` for an audit trail

Todo:
Enqueue the `StockPricesUpdateJob` once the `StockPurchaseJob` is completed


test/jobs/stock_purchase_job_test.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class StockPurchaseJobTest < ActiveJob::TestCase
+  test "it processes pending orders" do
+    pending_order1 = Order.create!(user: users(:one), stock: stocks(:one), status: "pending")
+    pending_order2 = Order.create!(user: users(:one), stock: stocks(:one), status: "pending")
+    completed_order = Order.create!(user: users(:one), stock: stocks(:one), status: "completed")
+
+    execute_calls = []
+
+    PurchaseStock.define_singleton_method(:execute) do |order|
```
jonny5, line 11: wasn't sure how to properly mock things out in minitest, this was what seemed to work?

app/services/purchase_stock.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def portfolio
+    @portfolio ||= order.portfolio
+  end
+
+  def withdrawal_amount
+    -order.purchase_cost
+  end
```
h-m-m, line 36: I like this a lot. The code is so legible

test/models/order_test.rb
```diff
@@ -10,4 +14,17 @@ class OrderTest < ActiveSupport::TestCase
...
+
+  test "can filter orders by pending status" do
+    pending_orders = [Order.create(stock: Stock.first, shares: 5, status: :pending, user: @student)]
+    Order.create(stock: Stock.first, shares: 5, status: :completed, user: @student)
+
+    assert_equal pending_orders, Order.pending
+  end
+
+  test "calculates purchase cost" do
+    stock = Stock.create(ticker: "EVG", price: 10.00)
```
h-m-m, line 26: There should already be a stock, defined in the fixtures folder, that exists when this test is run that can use instead of creating a new one

test/services/purchase_stock_test.rb
```diff
@@ -0,0 +1,48 @@
...
+  test "it creates an linked entry in portfolio_stocks" do
+    order = Order.create(stock: Stock.first, shares: 5, status: :pending, user: @student)
+    assert_difference("PortfolioStock.count") do
+      PurchaseStock.execute(order)
+    end
+    portfolio_stock = PortfolioStock.last
+    assert_equal portfolio_stock, order.portfolio_stock
+  end
+
+  test "it updates the order status to completed" do
```
h-m-m, line 31: Do we have time this weekend to test any failure states?

test/models/order_test.rb
jonny5: Updated to use fixture. Got it, I usually use FactoryBot for these types of tests

test/services/purchase_stock_test.rb
jonny5: Added a failure state at the bottom, required `mocha` gem for stubbing (seemed like a popular way in minitest? I tend to use rspec)

test/models/order_test.rb
h-m-m: I like FactoryBot. We may eventually want it on this project. I like the idea of using the fixtures for now until we have a compelling reason to switch

test/services/purchase_stock_test.rb
h-m-m: I wonder if there's a way we can do this without the stub you've written. I don't want to hold up the good work in this PR, though. If you're up for taking a go at rewriting it, I'm happy to pair with you to go over my concerns. If you're not up for that, then I think we should merge this PR in and we can discuss putting follow-up work in a separate issue

jonny5: I won't have availability for a bit so feel free to do what seems best. The mocking + failure test are contained in the last commit so you could remove that before merging if you don't want mocha added

LizPrescott: I've been noodling on this. For the kind of fail cases we want to test, are you talking about something like "what if there is not enough money in the account" or something like that @h-m-m ?

h-m-m: Yes! I think that's the main situation.

We will run into a problem at some point where the pending orders plus the new order could add up to more than what's in the account. Should that also go in this PR, or should we break that out into a follow-up PR?

LizPrescott: I'm glad you asked this because it made me realize some of this logic should maybe live in a more traditional controller request. Can't we process the transactions themselves based on the price the user sees when they make the request. I would think the only part we need to put in a chron job is updating the price of the stock, either at the beginning or end of the trade day. I'm thinking the purchases themselves should just use that number so the user can see a more immediate error message if they don't have enough in their account. Am I making sense? Is there another reason to put the transactions themselves in a job?

pollygee (approved):

PR #176, opened 2024-06-05
Changed checkbox partial to checkbox input in place of button

ZahBenjamin:


LizPrescott (approved):
Thanks for improving our accessibility! :)

PR #191, opened 2025-04-18
Add admin column to user and check for admin for access to admin pages

pollygee:
Added an `admin` column to Users table and used that to add a check for admin flag in order for a user to access admin pages.


maebeale (approved):
Looks good!

PR #212, opened 2025-04-26
Test school related models

AbuMareBear:
Fixtures are a great way to keep test setup simple and focused. They work especially well for establishing common baseline data that many tests can rely on.

However, fixtures shouldn't be used for every possible data permutation. Trying to model every variation with fixtures tends to make them bloated, hard to understand, and brittle. Similarly, mutating existing fixtures inside a test to fit a specific need is generally considered an anti-pattern — it undermines the clarity and reliability that fixtures are supposed to provide.

Because of these trade-offs, I recommend a hybrid approach:

Use fixtures for stable, shared baseline data that benefits multiple tests.
Use factories when a test needs a custom or dynamic setup that doesn't cleanly fit into a shared fixture.
This way, we get the best of both worlds: fast, consistent tests where possible, and flexible, explicit setup when needed.


test/controllers/classrooms_controller_test.rb
```diff
@@ -1,66 +1,73 @@
 require "test_helper"
 
 class ClassroomsControllerTest < ActionDispatch::IntegrationTest
-  setup do
-    @classroom = classrooms(:one)
-    @user = users(:one) # Assuming you have a user fixture named :one
-    sign_in @user # S
-  end
```
AbuMareBear, line 8: In my opinion, tests should be explicit. Having setup code like this makes it harder to understand. DRY should not apply in tests.

test/fixtures/classrooms.yml
```diff
@@ -1,13 +1,5 @@
...
-two:
-  name: MyString
-  year: two
-  school: two
-  grade: MyString
+hubbard_5th_grade:
+  name: Ms. Hubbard
+  year: current
+  school: armistead_elementary
+  grade: 5th
```
AbuMareBear, line 5: Fixtures should have realistic data.

PR #214, opened 2025-04-27
Test Year Model

AbuMareBear:
This PR:
- adds tests for the year model.
- makes the year column not nullable on the database level.
- uses realistic data for year fixture.
- adds the factory_bot_rails gem for factories.

### Fixtures and Factories
Fixtures are a great way to keep test setup simple and focused. They work especially well for establishing common baseline data that many tests can rely on.

However, fixtures shouldn't be used for every possible data permutation. Trying to model every variation with fixtures tends to make them bloated, hard to understand, and brittle. Similarly, mutating existing fixtures inside a test to fit a specific need is generally considered an anti-pattern that undermines the clarity and reliability that fixtures are supposed to provide.

Because of these trade-offs, I recommend a hybrid approach:
- Use fixtures for stable, shared baseline data that benefits multiple tests.
- Use factories when a test needs a custom or dynamic setup that doesn't cleanly fit into a shared fixture.

This way, we get the best of both worlds: fast, consistent tests where possible, and flexible, explicit setup when needed.

### DIFF
--- a/db/migrate/20250427103023_add_not_null_constraint_to_year_on_years.rb
+++ b/db/migrate/20250427103023_add_not_null_constraint_to_year_on_years.rb
@@ -1,0 +1,5 @@
+class AddNotNullConstraintToYearOnYears < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :years, :year, false
+  end
+end
--- a/test/factories.rb
+++ b/test/factories.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :year do
+    year { 2000 }
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ end
 
 group :test do
   gem "capybara"
+  gem "factory_bot_rails"
   gem "selenium-webdriver"
   gem "webmock"
   gem "simplecov", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -133,6 +133,11 @@ GEM
     docile (1.4.1)
     drb (2.2.1)
     erubi (1.13.1)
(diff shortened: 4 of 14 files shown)


db/migrate/20250427103023_add_not_null_constraint_to_year_on_years.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotNullConstraintToYearOnYears < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :years, :year, false
```
AbuMareBear, line 3: This is assuming that we do not have existing rows with null values or that we are concerned with Postgresql blocking reads and write, while it checks existing rows for `NULL` values.

test/factories.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :year do
+    year { 2000 }
```
AbuMareBear, line 3: This is the year that the idea of Stocks in the future was first pitched. 😎

danielabar: Assuming that `Year` is a model, might want to establish a pattern of one factory file per model. That will make it easier to find and maintain these as the project grows. Example from Human Essentials: https://github.com/rubyforgood/human-essentials/tree/main/spec/factories

pollygee: I like the idea of one factory file per model that organization is great as we get bigger!  I don't think it has to hold up this PR but is a good idea for future, what do you think @AbuMareBear?

pollygee (approved):
This is great @AbuMareBear thanks so much for adding Factory Bot!  Longer term we will probably want to move away from fixtures for simplicity but this is a really great start.

PR #233, opened 2025-05-23
Refactor schoolyear

pollygee:


Gemfile
```diff
@@ -24,14 +24,14 @@ group :development, :test do
...
+  gem "pry", "~> 0.15.2"
   # a linting tool, to encourage/enforce a consistent code style
   gem "standardrb"
   gem "standard-rails"
 end
 
 group :development do
   gem "bundler-audit"
-  gem "pry", "~> 0.15.2"
   # Use console on exceptions pages [https://github.com/rails/web-console]
```
maebeale, line 35: Why did you move this out of devt?

app/views/layouts/application.html.erb
```diff
@@ -1,9 +1,6 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <% if current_user&.admin? %>
-      <%= link_to "Admin Dashboard", admin_root_path %>
```
maebeale, line 5: Why isn't this needed anymore?

maebeale (approved):
Looks good, Polly!!!

You may want to additionally rename `Year#year` to either `year_value` or `name` so you don't have `year.year`, but that's nth and not required.

app/views/layouts/application.html.erb
maebeale: Makes total sense!

It was my bad -- I knew I reviewed before you were done.

pollygee:
Thanks @maebeale! Yes I agree re year.year. I’m not a fan.  The other thing I don’t like here is that we’re creating a school if one doesn’t exist based on how they type in a name of a school and year. This will cause multiples of the same school to get added based on misspells or capitalization. I think we should change that to a dropdown. 

I was thinking I’d creating separate issues for those two things rather than adding them here to keep things clean. Does that work @maebeale?

app/views/layouts/application.html.erb
pollygee: No not at all!  I appreciate you taking a look.

PR #241, opened 2025-05-25
Add unique email index and update user factory to prevent duplication.

yinho999:
- Added unique email index on user table
- update user factory to duplication email within tests

### DIFF
--- a/.standard_todo.yml
+++ b/.standard_todo.yml
@@ -1,4 +1,0 @@
-# Auto generated files with errors to ignore.
-# Remove from this list as you refactor files.
----
-ignore:
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -5,9 +5,10 @@ class UserTest < ActiveSupport::TestCase
     assert build(:user).validate!
   end
 
-  test "should not allow duplicate email addresses" do
+  test "validate uniqueness of email" do
     create(:user, email: "test@example.com")
     new_user = build(:user, email: "test@example.com")
+
     assert_not new_user.valid?
     assert_includes new_user.errors[:email], "has already been taken"
   end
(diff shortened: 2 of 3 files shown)


.standard_todo.yml
```diff
@@ -2,5 +2,3 @@
 # Remove from this list as you refactor files.
 ---
 ignore:
-- app/models/user.rb:
-  - Rails/UniqueValidationWithoutIndex
```
AbuMareBear, line 6: Could this whole file be deleted?

db/schema.rb
```diff
@@ -135,6 +135,7 @@
     t.bigint "classroom_id"
     t.boolean "admin", default: false
     t.index ["classroom_id"], name: "index_users_on_classroom_id"
+    t.index ["email"], name: "index_users_on_email", unique: true
```
AbuMareBear, line 138: Are we missing a migration file?

test/models/user_test.rb
```diff
@@ -4,4 +4,11 @@ class UserTest < ActiveSupport::TestCase
   test "factory" do
     assert build(:user).validate!
   end
+
+  test "should not allow duplicate email addresses" do
```
AbuMareBear, line 8: Just a suggestion but I find that removing words like `should` make test descriptions easier to parse. 

`test "validate uniqueness of email"`

test/models/user_test.rb
```diff
@@ -4,4 +4,11 @@ class UserTest < ActiveSupport::TestCase
...
+    create(:user, email: "test@example.com")
+    new_user = build(:user, email: "test@example.com")
+    assert_not new_user.valid?
+    assert_includes new_user.errors[:email], "has already been taken"
```
AbuMareBear, line 12: This is a good test. It could be a good idea to add a line between the test setup and the execution.

```
email = "test@example.com"
create(:user, email:)
user = build(:user, email:)

assert_not user.valid?
assert_includes user.errors[:email], "has already been taken"
```

AbuMareBear (commented):
Overall, this looks great.

AbuMareBear (approved):

PR #246, opened 2025-05-26
Update `Year` model to replace `year` attribute with `name`

### LINKED ISSUE #244: Fix Year model
Year represents a school year.  School years look like `2022 - 2023` and we won't need to do math on years, and therefore should be a string not an integer.

For this ticket, make Year a string type instead of integer, and change the year attribute to be `name` so we can use `year.name` instead of `year.year`.
Add seed data.  
They might want to add previous years, and we should add 10 years from now as well.  We want: 
2023 - 2024 
... through ...
2035 - 2036

yinho999:
- Replaced the `year` attribute with `name` across the `Year` model.
- Updated validations, migrations, related controllers, views, tests, and seeds to reflect the updated naming and data type.
Fixing #244

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,7 +8,7 @@
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
 
-year = Year.find_or_create_by(year: 2024)
+year = Year.find_or_create_by(name: "2024")
 
 school = School.find_or_create_by(name: "Test School")
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,6 +9,10 @@
 #   end
 
 year = Year.find_or_create_by(name: "2024")
+(2024...2036).each do |i|
+  year = Year.find_or_create_by(name: "#{i} - #{i + 1}")
+  year.save!
+end
 
 school = School.find_or_create_by(name: "Test School")
 
--- a/test/factories/years.rb
+++ b/test/factories/years.rb
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :year do
-    sequence(:year) { |n| 2000 + n }
+    sequence(:name) { |n| (2000 + n).to_s }
   end
 end
--- a/test/factories/years.rb
+++ b/test/factories/years.rb
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :year do
-    sequence(:name) { |n| (2000 + n).to_s }
+    sequence(:name) { |n| "#{200 + n} - #{2000 + n + 1}" }
   end
 end
--- a/app/controllers/classrooms_controller.rb
+++ b/app/controllers/classrooms_controller.rb
@@ -20,7 +20,7 @@ class ClassroomsController < ApplicationController
     @classroom = Classroom.new(classroom_params.except(:school_name, :year_value))
 
     school = School.find_or_create_by(name: classroom_params[:school_name])
-    year = Year.find_or_create_by(year: classroom_params[:year_value])
+    year = Year.find_or_create_by(name: classroom_params[:year_value])
     school_year = SchoolYear.find_or_create_by(school: school, year: year)
 
     @classroom.school_year = school_year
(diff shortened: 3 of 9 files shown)


db/seeds.rb
```diff
@@ -8,7 +8,7 @@
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
 
-year = Year.find_or_create_by(year: 2024)
+year = Year.find_or_create_by(name: "2024")
```
pollygee, line 11: A school year should look like "2024 - 2025" because in the US that's how we represent school years and this is the biggest reason we need it to be a string not an integer.
We are going to remove the ability to "create a year" from the UI so the naming will be consistent and we don't have a "2024-2025" and a "2024 - 2025" (spacing differences) in the database.

So I would love it here in the seeds if we had all the years to seed the database with our consistent spacing.

Could you add the previous few years and the next 10 years of years in school year format? If you're not able to do it at this time I can add it as a different PR.  If you could add this in though it would be really helpful.

What we need is: 
Year.find_or_create_by(name: "2023 - 2024")
Year.find_or_create_by(name: "2024 - 2025")
Year.find_or_create_by(name: "2025 - 2026")
Year.find_or_create_by(name: "2026 - 2027")
Year.find_or_create_by(name: "2027 - 2028")
Year.find_or_create_by(name: "2028 - 2029")
Year.find_or_create_by(name: "2029 - 2030")
Year.find_or_create_by(name: "2030 - 2031")
Year.find_or_create_by(name: "2031 - 2032")
Year.find_or_create_by(name: "2032 - 2033")
Year.find_or_create_by(name: "2033 - 2034")
Year.find_or_create_by(name: "2034 - 2035")

test/factories/years.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :year do
-    sequence(:year) { |n| 2000 + n }
+    sequence(:name) { |n| (2000 + n).to_s }
```
pollygee, line 3: I think we want our factory to look like a real school year as well so we could use some interpolation here and have `(2000 + n) -  (2000 + n+1)` or something like that so the string will represent what we should expect them to look like.

pollygee:
Thanks so much for this @yinho999 you're on a roll!!
I requested a few changes, mostly around us making it so our naming of years is more consistent.  If you're able to make those changes it would be really helpful.  Thank you!

db/seeds.rb
yinho999: > A school year should look like "2024 - 2025" because in the US that's how we represent school years and this is the biggest reason we need it to be a string not an integer.
> We are going to remove the ability to "create a year" from the UI so the naming will be consistent and we don't have a "2024-2025" and a "2024 - 2025" (spacing differences) in the database.
> 
> So I would love it here in the seeds if we had all the years to seed the database with our consistent spacing.
> 
> Could you add the previous few years and the next 10 years of years in school year format? If you're not able to do it at this time I can add it as a different PR.  If you could add this in though it would be really helpful.
> 
> What we need is: 
> ```Year.find_or_create_by(name: "2023 - 2024")
> Year.find_or_create_by(name: "2024 - 2025")
> Year.find_or_create_by(name: "2025 - 2026")
> Year.find_or_create_by(name: "2026 - 2027")
> Year.find_or_create_by(name: "2027 - 2028")
> Year.find_or_create_by(name: "2028 - 2029")
> Year.find_or_create_by(name: "2029 - 2030")
> Year.find_or_create_by(name: "2030 - 2031")
> Year.find_or_create_by(name: "2031 - 2032")
> Year.find_or_create_by(name: "2032 - 2033")
> Year.find_or_create_by(name: "2033 - 2034")
> Year.find_or_create_by(name: "2034 - 2035")```

Thanks for giving me some feedback. I previously misunderstood how the data should be look like. I will fix this issue soon.👍

pollygee: Oh great thanks so much!  Sorry if what I wrote before wasn't clear enough!  Thanks for working with me on it 😄

yinho999: @pollygee Thanks a lot for reviewing my PR again. Please let me know if I missed anything 🙏

pollygee (approved):
Thanks for making the additional updates @yinho999

PR #248, opened 2025-05-28
Fix #247: Add Teacher, Student, Admin seeds, update README

### LINKED ISSUE #247: Update User seed data
We currently create a user in the seed data.  We need to update this so we create a user of each type.  

Create a Teacher a Student and an Admin.
The name for each of these should be the role name so we can differentiate them.
The password for these should each should be the same and something simple like "password".

Please also update the Readme so that new folks will know how to utilize the seed data right away upon getting the app running locally, or for when we have the app deployed.

nizam12khan:
Issue: Resolves #247

Changes:

- Updated db/seeds.rb to create three users (Teacher, Student, Admin) with usernames matching their roles, emails, and password "password". Used admin: true for the Admin user (no role column exists).

- Ensured idempotency in db/seeds.rb by clearing existing users.

- Updated README.md with seed data instructions, specifying username-based login and credentials.

Testing:

- Ran bin/rails db:seed and verified users were created with correct usernames and admin flags.

- Logged in as each user (Teacher, Student, Admin) using username and password.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,19 +36,17 @@ It is **strongly** recommend to use Docker. See instructions above.
 
 ## Seed Data
 
-The application includes seed data to create default users for testing. After setting up the database, you can seed it with: `bin/rails db:seed`
+After running `bin/rails db:setup`, the database will automatically be seeded with three default users.
 
-This creates three users:
+| Role    | Username | Email                  | Password |
+|---------|----------|------------------------|----------|
+| Teacher | Teacher  | teacher@example.com    | password |
+| Student | Student  | student@example.com    | password |
+| Admin   | Admin    | admin@example.com      | password |
 
-Teacher: Username: Teacher, Password: password (Email: teacher@example.com)
+Use the **username** and **password** to log in and test the application locally.
 
-Student: Username: Student, Password: password (Email: student@example.com)
-
-Admin: Username: Admin, Password: password (Email: admin@example.com)
-
-Use the username and password to log in and test the application locally or on a deployed environment. 
-
-Note: Do not run db:seed in production unless explicitly required.
+> ⚠️ **Note:** Do not run `db:seed` in production unless explicitly required.
 
 ## URL
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -24,7 +24,7 @@ classroom = Classroom.find_or_create_by(name: "Smith's Sixth Grade", school_year
 User.destroy_all
 
 # Create users with usernames and admin flag
-User.create!(
+Teacher.create!(
   username: "Teacher",
   email: "teacher@example.com",
   password: "password",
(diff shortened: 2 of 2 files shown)


nizam12khan:
Hi @pollygee and @americac ,
I’ve completed the changes for issue #247 and would appreciate your feedback! I have a couple of questions:

README Updates: I’ve added a "Seed Data" section in the README with the user credentials and login instructions. Are there any additional details or changes you’d recommend for the README to make it more helpful for users?

Refactoring db/seeds.rb: Currently, I’m using User.create! three times to create the Teacher, Student, and Admin users in db/seeds.rb. Would it be better to refactor this into a helper method (e.g., create_user) to make the code DRYer, or is the current approach fine as is?

Looking forward to your suggestions! Thanks for reviewing.

README.md
```diff
@@ -34,6 +34,22 @@ $ docker compose run stocks rails test
...
+Teacher: Username: Teacher, Password: password (Email: teacher@example.com)
+
+Student: Username: Student, Password: password (Email: student@example.com)
+
+Admin: Username: Admin, Password: password (Email: admin@example.com)
+
+Use the username and password to log in and test the application locally or on a deployed environment. 
+
+Note: Do not run db:seed in production unless explicitly required.
+
```
pollygee, line 52: The README updates look great thanks!

db/seeds.rb
```diff
@@ -20,10 +20,35 @@
...
+User.create!(
+  username: "Admin",
+  email: "admin@example.com",
+  password: "password",
+  password_confirmation: "password",
+  admin: true,
+  classroom: classroom
+)
+
+Rails.logger.info "Seeded 3 users: Teacher, Student, and Admin"
```
pollygee, line 54: Thanks for adding the logger 👍

db/seeds.rb
```diff
@@ -20,10 +20,35 @@
...
-  user.password = "password"
-  user.password_confirmation = "password"
-  user.classroom = classroom
-  user.save!
-end
+# Clear existing users to ensure idempotency
+User.destroy_all
+
+# Create users with usernames and admin flag
+User.create!(
```
pollygee, line 27: Is there any reason we wouldn't want to use `Teacher.create!` to create the teacher here?
I think it would be good if we used it instead of `User.create` to be explicit about what we're creating.  It also gives us the ability to use things like `is_a?(Teacher)`

db/seeds.rb
```diff
@@ -20,10 +20,35 @@
...
+  username: "Teacher",
+  email: "teacher@example.com",
+  password: "password",
+  password_confirmation: "password",
+  admin: false,
+  classroom: classroom
+)
+
+User.create!(
```
pollygee, line 36: Same here.  I think we could use `Student.create!`

README.md
nizam12khan: image
Made few changes to Readme, need your comments

pollygee (approved):
Wonderful!  Thanks for making the small updates.  Really appreciate your help.

PR #255, opened 2025-05-30
Fixes #249 Add stocks route and controller 

### LINKED ISSUE #249: Add Stocks Route and controller
Teachers and students should be able to view stocks.
Let's add Stocks routes, we only need index and show as they won't be able to create edit or destroy stocks.

Create the Controllers and views for these routes as well. 

Index view can just be a simple list of stocks (similar to what we have for classrooms or schools)
Show view can just show the fields we have on a stock.

yinho999:
Fixes #249 
Added
- seed stock data
- stock view
- stock controller
- controller testing for both view and json format


yinho999:
Hello @pollygee. I've updated the stocks UI layout to better handle and display extensive stock data lists. Please let me know if this approach works well or we can revert back to the original design.

From
image

To
image

app/views/stocks/show.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.partial! "stocks/stock", stock: @stock
```
AbuMareBear, line 1: Can we remove this?

db/seeds.rb
```diff
@@ -16,6 +16,27 @@
 
 school = School.find_or_create_by(name: "Test School")
 
+Stock.find_or_create_by(ticker: "AAPL",
```
AbuMareBear, line 19: This might be more readable if every key is on it's own line.

db/seeds.rb
```diff
@@ -16,6 +16,27 @@
...
+  stock_exchange: "NASDAQ",
+  company_name: "Apple Inc.",
+  company_website: "https://www.apple.com",
+  description: "Apple Inc. specializes in the conceptualization, production, and distribution of smartphones," \
+  " personal computers, tablets, wearable technology, accessories, and various associated services. " \
```
AbuMareBear, line 24: Question: Do we need a long description like this? Might be more readable to have a shorter one line description if the requirements allow for it.

test/controllers/stocks_controller_test.rb
```diff
@@ -0,0 +1,40 @@
+require "test_helper"
+
+class StocksControllerTest < ActionDispatch::IntegrationTest
+  def setup
+    @stock = create(:stock)
+    @user = create(:user)
+    sign_in @user
+  end
```
AbuMareBear, line 8: It has been my experience that having test set up like this makes tests harder to write and maintain. In my opinion every test should have its own setup. DRY (Don't Repeat Yourself) doesn't apply to testing.

app/views/stocks/_stock.json.jbuilder
```diff
@@ -0,0 +1,2 @@
+json.extract! stock, :id, :ticker, :stock_exchange, :company_name, :company_website, :created_at, :updated_at
```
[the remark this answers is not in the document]
yinho999: I saw there is json version for school view as well, are we not going to use the json format?

db/seeds.rb
yinho999: I am not entirely sure about the structure of stocks. I can make it shorter if needed

app/views/stocks/_stock.json.jbuilder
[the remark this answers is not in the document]
AbuMareBear: Yeah. The Rails generator creates them but If we don't need JSON responses it's best not to have to support them. I've removed some other JSON views in other PRs.

app/helpers/stocks_helper.rb
```diff
@@ -0,0 +1,2 @@
+module StocksHelper
```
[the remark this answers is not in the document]
yinho999: There are other similar blank helpers in the same directory as well, should I remove those as well?

db/seeds.rb
pollygee: For now this is just for development so a short description (does not have to be accurate) would work just fine.

yinho999:
@pollygee @AbuMareBear 

Thank you for reviewing my PR and giving such detailed responses. I have reverted the styles of the stock page and removed the JSON file and related tests, as we currently do not need them.

app/controllers/stocks_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class StocksController < ApplicationController
+  before_action :set_stock, only: %i[show]
+  before_action :authenticate_user!
+
+  # GET /stocks or /stocks.json
```
AbuMareBear, line 7: Might be good to remove these comments.

test/controllers/stocks_controller_test.rb
```diff
@@ -0,0 +1,19 @@
+require "test_helper"
+
+class StocksControllerTest < ActionDispatch::IntegrationTest
+  test "should get index" do
+    @stock = create(:stock)
+    @user = create(:user)
+    sign_in @user
+    get stocks_url
+    assert_response :success
+  end
```
AbuMareBear, line 10: Does using local variables work here? Also might be good to add spaces between the stages of the test for readability.

```
stock = create(:stock)
user = create(:user)
sign_in user

get stocks_url

assert_response :success
```

test/factories/stocks.rb
```diff
@@ -1,4 +1,22 @@
...
+    sequence(:competitor_names) { |n| "Competitor #{n}, Competitor #{n + 1}" }
+    sales_growth { 5.5 }
+    industry_avg_sales_growth { 4.2 }
+    debt_to_equity { 0.3 }
+    industry_avg_debt_to_equity { 0.4 }
+    profit_margin { 15.2 }
+    industry_avg_profit_margin { 12.8 }
+    cash_flow { 1_000_000.00 }
+    debt { 500_000.00 }
+    price_cents { 12_500 } # $125.00
```
AbuMareBear, line 20: A best practice for factories is to only include required fields. A great task would be to remove any optional fields here and add validations in the stock model for attributes that are not optional. That could be a separate PR.

PR #266, opened 2025-06-05
Add delete link to classrooms list with turbo method

### LINKED ISSUE #252: Add ability to delete a classroom on the classrooms index page
Right now the classrooms index page has show and edit buttons but you should also be able to delete a classroom here, so please add a delete button 
Right now:
Image

Add a delete button:
Image

yinho999:
Fix #252 
- Added delete link to the classroom index page
image


pollygee (approved):
Great thanks!

Issue #263, opened 2025-06-05
Style Login Page


pollygee:
Add styling to login page 
Use the [Simple Sign in styling from tailwinds](https://tailwindcss.com/plus/ui-blocks/application-ui/forms/sign-in-forms#component-766a0bf1b8800d383b6c5b77ef9c626c)

Should look like the following.  Logo is in the assets folder.  Button color is #00698C

We may remove "sign up for account" link at some point but don't worry about doing that now.  We need to think a lot more about how we will add users.

![Image](https://github.com/user-attachments/assets/87810d48-0190-4ca1-9153-4f8330ac0817)

mgrigoriev8109:
I'd like to work on this. Would you mind assigning me to the ticket?

PR #268, opened 2025-06-06
Fix login page styles 

mgrigoriev8109:
This PR should resolve [Style Login Page](https://github.com/rubyforgood/stocks-in-the-future/issues/263) by:
- Adding in the cap logo
- Replacing the username / password "labels" with "placeholders"
- copying over some base tailwind styles and tweaking them a bit

Test manually by running locally, and navigating to the login screen. Should look like this, and still function properly.

Screenshot 2025-06-06 at 2 07 21 PM

Notes: 

1. I'm seeing some errors in the console with the app attempting to import `shadcn.css`, is this known or something we should look into? 
2. I didn't any tests, as there currently aren't any around the login flow, and feel like if there need to be more discussions around login then those tests should be added at a later point.


pollygee:
@mgrigoriev8109 thanks for this.  

I didn't know about any problems with `shadcn.css`.  We recently upgraded to Rails 8 so it might be a fallout of that.  @abachman are you aware of anything outstanding?  Either way we can create a separate ticket to look into that!

I think it's fine for us to add tests later.  We won't have students going through this flow so we will def need to have some discussions around how to handle this.  We also likely won't want a signup link either.

I see that the navbar is showing up here and it shouldn't.  I can add a separate ticket to take care of that as well.

pollygee (approved):
Looks great!  thank you

PR #272, opened 2025-06-08
Add separate authentication for students

pollygee:
## This PR adds authentication for Student.
Students will not have emails so we need to user usernames to authenticate and we can not send passwords so teachers will tell passwords to students in person in the classroom.

Update validations to allow students to have a nil email.
Update student table to have `reset password` and `destroy links`

## Authentication flow:
```mermaid
graph TD
    A[User visits site] --> B[Single Login Form]
    B --> C{Authentication}
    C -->|Success| D{Check user.type}
    D -->|Student| E[Student Portfolio]
    D -->|Teacher| F[Classrooms index Page]
    D -->|Admin| F[Classrooms index Page]
    C -->|Failure| H[Login Error]
    H --> B
```
### New view of the Classroom page with updates to the student table:
StocksInTheFuture

## In a future PR we will 
-  add ability to bulk upload students.
- clean up permissions and make sure that students only see what they should see
- remove email from student, but first we should check that we will never want to use it


app/controllers/classrooms_controller.rb
```diff
@@ -1,12 +1,16 @@
 class ClassroomsController < ApplicationController
   before_action :set_classroom, only: %i[show edit update destroy]
   before_action :authenticate_user!
+  before_action :ensure_teacher_or_admin, except: [:index, :show]
```
maebeale, line 4: Students can see classrooms index and classroom show?

maebeale: Nvm, I see it below now

app/controllers/classrooms_controller.rb
```diff
@@ -61,4 +65,20 @@ def set_classroom
   def classroom_params
     params.require(:classroom).permit(:name, :grade, :school_name, :year_value)
   end
+
+  def ensure_teacher_or_admin
```
maebeale, line 69: No required, but you might want to consider making this a controller concern since you have it in more than one. Or, adding it to ApplicationController

maebeale (approved):
Looks good! It's a big PR and a big leap forward in functionality!

PR #275, opened 2025-06-10
Rubocop violations in a TODO file so we andle them and not allow more

pollygee:
Ran rubocop -A for cleanup

Put all rubocop errors in TODO file so we can have a clean rubocop run and not allow new errors
I'll add an issue so we can start tackling the errors in the todo file


pollygee:
Oh no!  You're right @AbuMareBear thanks for looking out!!

AbuMareBear:
@pollygee No worries. I did the same thing!

PR #307, opened 2025-07-01
[302] Install strong migrations gem

jwshinx:
To avoid unsafe migrations, this installs the `strong_migrationss` gem. When you run a migration that’s potentially dangerous, you’ll see an error message like:

```
=== Dangerous operation detected #strong_migrations ===

Active Record caches attributes, which causes problems
when removing columns. Be sure to ignore the column:

class User < ApplicationRecord
  self.ignored_columns += ["name"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class RemoveColumn < ActiveRecord::Migration[8.0]
  def change
    safety_assured { remove_column :users, :name }
  end
end
```

An operation is classified as dangerous if it either:
- Blocks reads or writes for more than a few seconds (after a lock is acquired)
- Has a good chance of causing application errors

See [docs for all the ways it helps you write safe migrations](https://github.com/ankane/strong_migrations).


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Mark existing migrations as safe
+# rubocop:disable Style/NumericLiterals
+StrongMigrations.start_after = 20250701200226
+# rubocop:enable Style/NumericLiterals
```
jwshinx, line 6: Because `20_250_701_200_226` doesnt make this timestamp more readable

pollygee (approved):
👍 thank you!

PR #310, opened 2025-07-01
[Issue-309] Remove transactions column of the Portfolios table

jwshinx:
This column is redundant and we don't need it. We have `portfolio_transactions` to store structured data. Here, we remove the column and associated application level code.


pollygee (approved):
Looks great thanks!

PR #322, opened 2025-07-02
Make sure new students get an empty portfolio

### LINKED ISSUE #304: Update Portfolio to be required
Students should be required to have a porfolio. 

We want to create an empty portfolio when a student is created.
We will need to add a portfolio for the students currently in our db.

Portfolio should be nested under user.

pollygee:
Make sure there is a portfolio creation for new students

Add `accepts_nested_attributes_for` Portfolio to User model to enable portfolio creation via nested attributes

Add `after_create`callback to Student model to guarantee every student gets a portfolio

Update seeds.rb to give them a portfolio

Remove controller-specific portfolio creation because the logic is now in the  model

Resolves #304

### DIFF
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -78,5 +78,4 @@ class StudentsController < ApplicationController
     numbers = (1..99).to_a
     "#{words.sample}#{numbers.sample}"
   end
-
 end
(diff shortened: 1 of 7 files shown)


dpowers1 (approved):
Seems good!

PR #325, opened 2025-07-03
[Issue 320] Enable successful student login to bring up their dashboard with user-portfolio endpoint

jwshinx:
This pr implements the user-portfolio nested style `http://localhost:3000/users/2/portfolios/1` path on successful login.

Additionally it includes docker changes to enable running tests in a container.

~~The create-student-user form needs work; so in lieu, I quick-and-dirty added the callback to automatically create a portfolio. Handy for console, for now.~~

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -72,3 +72,72 @@ User.create!(
 )
 
 Rails.logger.info "Seeded 3 users: Teacher, Student, and Admin"
+
+mike = Student.create!(
+  username: "mike",
+  email: "mike@example.com",
+  password: "password",
+  password_confirmation: "password",
+  admin: false,
+  classroom: classroom,
+  portfolio_attributes: { current_position: 10_000.0 }
+)
+portfolio = mike.portfolio
+stock = Stock.first
+
+######################################################################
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -7,6 +7,8 @@ class User < ApplicationRecord
   accepts_nested_attributes_for :portfolio
   has_many :orders, dependent: :destroy
 
+  after_create :create_portfolio_for_student
```
pollygee, line 10: We have this method on [Student](https://github.com/rubyforgood/stocks-in-the-future/blob/main/app/models/student.rb#L6).  Do you think it should be here instead of on Student?

enhanced-student-authentication-implementation-plan.md
```diff
@@ -64,7 +64,7 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path_for(resource)
```
pollygee, line 64: I think we want to remove this file 😆 .  I think I accidentally introduced it but it was a plan I created and shared with someone before doing work lol.

jwshinx: ok, ill remove...

PR #326, opened 2025-07-03
Updating seeds file 

warlita:
Added the following stocks to the seeds file:
Coca Cola
Disney
Verizon
Southwest
Under Armour
Gap
Ford
Sony
Sirius XM


pollygee (approved):
This looks good @warlita!  Well done! 🥳 Thanks for your work on this!
The test.rb file is empty so we don't need it right now, it's better not to add things we don't need.
Can you delete that before merging?  I'll approve your PR so you're cleared to merge whenever you delete the extra file!

PR #331, opened 2025-07-03
Add relationship for teacher to have multiple classrooms

### LINKED ISSUE #316: Teachers should be able to have multiple classrooms

dpowers1:
Closes #316

### DIFF
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 
 class Teacher < User
+  has_many :teacher_classrooms, dependent: :destroy
+  has_many :classrooms, through: :teacher_classrooms
 end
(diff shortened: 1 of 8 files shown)


app/models/teacher.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 
 class Teacher < User
+  has_many :teacher_classrooms, dependent: :destroy
```
pollygee, line 4: Do we want to destroy the classroom if we remove the teacher?  We still have students who are associated with it right?  maybe it's fine because the classroom itself isn't that important.

dpowers1: This will just destroy the association between a teacher and classroom is a teacher is deleted. The classroom should still stay

PR #335, opened 2025-07-03
Issue 330: Add grade books

spaulsandhu:
### Overview

This adds the basic infrastructure to allow us to start building out the Grade Book feature. I've introduced some new models here, please read each commit message individually for more context about what each model is doing.

Note: This is still very early days and things will probably change here pretty fast, but I wanted to get the basic skeleton in place for me to start building out the robust UX for this feature. I'm sure I'm going to find edge cases and incorrect assumptions in the data model as I work through the next steps lol.


app/models/quarter.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  scope :ordered, -> { order(:number) }
+
+  def next
+    school_year.quarters.ordered.find_by(number: number + 1)
+  end
+
+  def previous
+    school_year.quarters.ordered.find_by(number: number - 1)
+  end
```
spaulsandhu, line 20: I'm going to revisit this in future work, this is more experimental at the moment.

pollygee: I think this will work for a V1 anyway.  I don't think we need to worry about a previous quarter from a previous year but I'll add it to my list to check on.

PR #336, opened 2025-07-03
issue-332 add standard button styling

jwshinx:
We want a solution that everyone knows. [ViewComponent](https://github.com/ViewComponent/view_component) seems like overkill.

Just some light css. organized under the `/custom` subdir. Not sure about how we want to name classes.

Pull the branch, rebuild and log in as a student. Go to their trading floor and the "buy" "sell" buttons should look exactly the same.

This is just laying a seam. I expect follow up `*.css` files for `inputs` or `forms`, etc.

## Consideration
- for some reason, `shadow-xs` doesnt play nice. so i removed it. its a tw standard, but why? ill need to dig

### DIFF
--- a/app/assets/stylesheets/custom/buttons.css
+++ b/app/assets/stylesheets/custom/buttons.css
@@ -1,0 +1,7 @@
+.tw-btn-primary {
+  @apply rounded-md bg-[#00698c] px-4 py-2 text-sm font-semibold text-white hover:bg-[#004f6b] focus-visible:outline-2 focus-visible:outline-offset-2;
+}
+
+.tw-btn-secondary {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/custom/buttons.css
```diff
@@ -0,0 +1,7 @@
+.tw-btn-primary {
+  @apply rounded-md bg-[#00698c] px-4 py-2 text-sm font-semibold text-white hover:bg-[#004f6b] focus-visible:outline-2 focus-visible:outline-offset-2;
```
dpowers1, line 2: I think there is a tailwind way to make custom css variables in a @theme variable but it can be a follow up 

https://tailwindcss.com/docs/colors#customizing-your-colors

dpowers1 (approved):
Looks great!

PR #344, opened 2025-07-03
318 buy stock page

### LINKED ISSUE #318: Create Buy Stock page
Create Buy Stock page. Make this an individual page, will be turned into a modal or drawer later
https://www.figma.com/board/5Enzwl1rJVYdXFsaCfcJmI/Stocks-In-the-Future?node-id=754-1419&t=rixBrqK3aZPlr04t-4

americac:
Resolves #318 & #319


app/views/orders/new.html.erb
```diff
@@ -1,9 +1,3 @@
-<h1>New order</h1>
-
-<%= render "form", order: @order %>
-
-<br>
-
-<div>
-  <%= link_to "Back to orders", orders_path %>
```
pollygee, line 8: Do we not want this link anymore?

app/controllers/orders_controller.rb
```diff
@@ -2,29 +2,23 @@
...
-  # GET /orders/1/edit
   def edit; end
 
-  # POST /orders or /orders.json
   def create
-    @order = Order.new(order_params)
-
-    @order.user = current_user
-    @order.stock = Stock.find(order_params[:stock_id])
+    @order = Order.new(order_params.merge(user: current_user))
```
pollygee, line 21: Do we want to restrict this here to make sure users can only create orders on their own portfolio? Or we can add this as a separate ticket to make sure we follow up on this if you agree we need it but you don't want to add it now.

app/models/order.rb
```diff
@@ -17,4 +23,19 @@ class Order < ApplicationRecord
   def purchase_cost
     stock.price_cents * shares
   end
+
+  private
+
+  def create_portfolio_transaction
+    PortfolioTransaction.create!(
```
pollygee, line 30: We prob need a follow up ticket here to make sure that the portfolio has sufficient funds before creating the transaction.

pollygee (approved):
This looks great!  I see some places for improvements we could just note and put into new tickets for follow ups.  Thanks so much!

app/views/orders/new.html.erb
americac: It was not part of the design so I removed it but I can certainly re-add it.

pollygee: Oh I see.  Because it was supposed to be a modal we wouldn't have needed a link to go back but since it's a page we prob need a way to navigate back to the orders index.  You could re-add it or we can add it back in with another issue.

PR #349, opened 2025-07-03
[Issue-332] more consolidation of buttons styles; also forms.css for input element styles

jwshinx:
We want to encourage tailwind utility class changes to be streamlined -- in one place. That way, it just cascades to all the view elements automatically.
- More button work
- Css file to consolidate form element style classes

## Buttons
We do that here for buttons -- they are centralized as css classes in `app/assets/stylesheets/custom/buttons.css`. I did not swap out all partials with direct tw utility class markups. That will come later, when those flows are worked on.

The focus here is consolidation. We expect change, eg `tw-btn-buy` and `tw-btn-sell` were simply created to accommodate the likelihood that they are different from the rest.

## Forms
Same idea with form elements.

Screenshot 2025-07-03 at 2 17 27 PM

Screenshot 2025-07-03 at 4 01 15 PM

### DIFF
--- a/app/assets/stylesheets/application.tailwind.css
+++ b/app/assets/stylesheets/application.tailwind.css
@@ -2,6 +2,7 @@
 
 /* SIF custom CSS */
 @import 'custom/buttons.css';
+@import 'custom/forms.css';
 
 @tailwind base;
 @tailwind components;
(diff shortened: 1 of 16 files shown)


spaulsandhu (approved):
Love it!!! 👍🏽

pollygee:
Amazing!  Thank you.

PR #408, opened 2025-08-21
342 and 406 grade book skeleton

spaulsandhu:
### Overview

This implements some of the skeleton views for the Grade Book feature and also includes some changes to the underlying data model for Grade Entries. Please view each commit separately for additional context on each change. This does NOT wire up any kind of interactivity for updating Grade Entry values, that will be coming up in a separate slice of work.

Screenshots of before/after for styling and the the basic Grade Book table:

Before styling changes you can see the table views were getting smushed:

Screenshot 2025-08-21 at 7 20 46 AM

After styling changes + skeleton view of Grade Books:

Screenshot 2025-08-21 at 7 20 10 AM
Screenshot 2025-08-21 at 7 20 19 AM


app/views/layouts/application.html.erb
```diff
@@ -8,12 +8,13 @@
     <%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>
     <%= javascript_importmap_tags %>
   </head>
-  <body class="h-screen w-screen inline-flex overflow-auto">
+  <body class="h-screen w-screen overflow-hidden">
```
spaulsandhu, line 11: I'm not exactly awesome with CSS so I fully anticipate some of this to change.

db/migrate/20250821130409_add_attendance_days_to_grade_entries.rb
```diff
@@ -0,0 +1,5 @@
+class AddAttendanceDaysToGradeEntries < ActiveRecord::Migration[8.0]
+  def change
+    add_column :grade_entries, :attendance_days, :bigint
```
spaulsandhu, line 3: Per the client's request, we'll change the model to support the attendance days values.

app/controllers/grade_books_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# app/controllers/grade_books_controller.rb
+class GradeBooksController < ApplicationController
+  def show
+    @classroom = Classroom.find(params[:classroom_id])
+    @grade_book = @classroom.grade_books.find(params[:id])
+  end
+end
```
spaulsandhu, line 9: Temporary dummy controller, I'll add in authorization layers in future work.

config/routes.rb
```diff
@@ -14,6 +14,7 @@
   end
 
   resources :classrooms do
+    resources :grade_books, only: :show
```
pollygee, line 17: should we also add this as a resource under admin at some point so we can add/manage them?

pollygee (approved):
This looks great!  Thanks a bunch. 

Looks like there's a conflict and you need to run Rubocop but when those are done I'm 👍 for a merge!

config/routes.rb
spaulsandhu: I haven't given this a lot of thought (I really don't have the context to say with any certainty), I think it would be good to discuss some of the business requirements here and come up with a plan for sure though.

PR #411, opened 2025-08-21
Allow Students and Teachers to be Added to Classrooms Separately

### LINKED ISSUE #391: Add ability to add Teacher and students separately on the admin Classroom page
currently on the admin classroom page we can add Users

we need to split this out so we can add Teachers and Students separately

Admin pages is using Administrate so we need to work around this for this 

Image

juliannatetreault:
Resolves #391  

### Description
Previously, students and teachers had to be added to a classroom via the "Users" field. This PR addresses this by allowing students and teachers to be added separately on the "Edit Classroom" page via the Admin Classroom page. This PR also adds tests for the changes.

### Testing Instructions
- Navigate to the Admin Classroom Edit page and select a classroom to be edited.
- Ensure that you are able to add students and teachers to the classroom, and that the fields have been separated. 
- Ensure that the students and teachers are displayed properly on the show page as well.

#### After: The Admin Classroom Edit Page 
Screenshot 2025-08-21 at 1 23 50 PM

#### After: The Admin Classroom Show Page 
Screenshot 2025-08-21 at 1 23 55 PM


test/models/classroom_test.rb
```diff
@@ -45,4 +45,47 @@ class ClassroomTest < ActiveSupport::TestCase
...
+    student2 = create(:student, classroom: classroom)
+    
+    assert_includes classroom.students, student1
+    assert_includes classroom.students, student2
+    assert_equal 2, classroom.students.count
+  end
+  
+  test "has many teachers through teacher_classrooms association" do
+    classroom = create(:classroom)
+    teacher1 = create(:teacher)
```
pollygee, line 61: Each of the teacher creations here will create a classroom then you're adding them to a different classroom.  I'm not sure it's necessary for the teachers to belong to two classrooms for this test.  Maybe we can just create the teacher with the classroom since you've already created it, like you did on line 51.

pollygee (changes requested):
Thank you so much for your first PR!! 

It looks like you need to run `Rubocop -A` and make sure sure that the errors are fixed there's currently some errors.

I also made a few small change requests.

test/models/classroom_test.rb
juliannatetreault: Good callout! I've updated this so that we're more closely mirroring the object setup in the student test, but doing it in such a way that appropriately sets up the `teacher_classroom` association.

I've also updated the linting in the test file 😊

Issue #415, opened 2025-08-23
Bug: Edit order page is showing as sell page


pollygee:
Steps:
Buy a stock
Click edit on the order show page

BUG: it takes you to the sell page for the stock, we need an edit page.  It's the right route I think it's prob just using the wrong partial

We need a page similar to the sell page but it should populate with how many stock there were from the order and the button should say "update" instead of "buy"

![Image](https://github.com/user-attachments/assets/03ac3e50-9f37-4278-a6d1-e862847587d1)

suraj-mandal:
One thing, I have also found is that, when an order is placed, it does not update the `portfolio_stock` table, and henceforth, I can see that even if the orders are placed, the stocks are not added to the portfolio, especially for the **completed** orders.

As you can see here:

<img width="1692" height="339" alt="Image" src="https://github.com/user-attachments/assets/dd66535b-58a5-456a-9ba5-74b04312ecc8" />

suraj-mandal:
Ideally, only pending orders should be editable, I think. Do let me know about this as well

suraj-mandal:
Another thing I want to bring to your attention is the fact that for all the orders, the transaction_type is coming as **null**; henceforth, even for **buy** orders, you are getting the **sell order** page, because the partial works based on the transaction type, which is coming as null.

It is verified by this statement:

```ruby
latest_order = Order.order(:created_at).reverse.first
latest_order.transaction_type
```

<img width="1676" height="307" alt="Image" src="https://github.com/user-attachments/assets/97baeced-c04a-4179-9158-4fe2c5e9be34" />

suraj-mandal:
During the creation of the order, the `transaction_type` gets persisted only in the memory during that time, and based on the `transaction_type`, the `portfolio_transaction` entry is being created. However, when you edit a **pending** order, this **transaction type** does not persist anymore, so you get the value as nil, as you do not have the **transaction type** stored in the memory.

There are two ways to mitigate this:

1. Change the order schema to include the transaction type
2. Get the transaction type from the Portfolio Transaction schema.

I am going with the second one.

pollygee:
> One thing, I have also found is that, when an order is placed, it does not update the `portfolio_stock` table, and henceforth, I can see that even if the orders are placed, the stocks are not added to the portfolio, especially for the **completed** orders.
> 
> As you can see here:
[...]

Issue #403, opened 2025-08-21
Holdings not matching from portfolio page to trading floor


pollygee:
The number of stocks owned of each stock should match on the portfolio page and trading floor page
they don't match

here's an example from staging.  They should both be doing the same calculation

![Image](https://github.com/user-attachments/assets/17884b6e-ed1a-439c-a66e-d39cc44642a5)

![Image](https://github.com/user-attachments/assets/54daf8a2-0cbc-4206-837d-072f121b8c0a)

davidpgero:
@pollygee I'd like to work on this one if it's free. 

And yes, I'm sorry. It's my fault. I had an idea, so I created a draft PR before asking. 

Next time I'll be smarter :)

pollygee:
Yeah @davidpgero I think this has been resolved.  Something went wrong with our automations  because I see that PRs  are linked here.

Sorry about that!  I'm trying to keep up but things are moving fast!  I can add you to the project since you're picking up issues on a fairly regular basis that way you can add yourself to any issue.  

The idea that you had did it fix something related to this that still isn't fixed? 

I actually think the problem here  from this ticket is that the profile page is just listing the transactions not the holdings.  So it really isn't a bug it's just a flaw in what previous us put on the page 😆 It doesn't make sense for it to be a list of transactions so we'll have to change that.  The designer is currently working on what will need to go on the page.  SO we'll have new tickets for that soon.

pollygee:
THanks!  I've sent you an invite for the project so you can self assign.  
You've been really helpful and super involved, if you're up for it - it could be helpful if you joined our slack channel in the ruby for good slack.  You can find the invite to slack on our website rubyforgood.org through the join us tab.

As far as those issues go I'm not sure what you mean by the "weighted average".  Do you mean the current value of the portolio as in 2 stocks * current_price + 3 stocks * other current_price etc? to get a total of the current value of the porfolio if the student was to sell everything at the current price?

> Unify access that we can use on the trading floor and portfolios as well
I'm not sure what you mean by "unify access" here.

> Portfolio page shows the current price, average price, and buy + sell working buttons
The designer is currently working on making some decisions about what should show on the portfolio page.  If you join the slack I'd be happy to share the figjam we're working in so you can see the decisions we're making/you can share ideas as we're making those calls.  But student portfolio would have buy sell buttons only for the stocks they currently hold.  That's where we're at now.

PR #423, opened 2025-08-24
Fixes Issue #415

suraj-mandal:
### The given PR addresses the following changes

1. Each pending order can be edited based on whether it is a buy order or a sell order. *(Currently only buy orders can be tested, sell orders cannot be tested due to issue [#403](https://github.com/rubyforgood/stocks-in-the-future/issues/403))*
2. The portfolio transaction for the pending order gets updated based on the number of shares updated
3. Added tests for both the updation of buy and sell pending orders.
4. Only pending orders can be updated. *(Do let me know if this is what you intend to do)*

### Updated Screenshots of the home page

Screenshot From 2025-08-25 01-17-21

### Updated Screenshots of the edit page

1. Buy order
image

### Results of **rake test**

Screenshot From 2025-08-25 01-10-40

### Extra Issue identified when testing

1. The new order button fails because of **null** stock_id. It placement should be changed, or it should be removed altogether.
Screenshot From 2025-08-25 01-08-48


app/models/order.rb
```diff
@@ -51,6 +67,28 @@ def sufficient_funds_for_buy
     errors.add(:shares, "Insufficient funds. You have #{formatted_balance} but need #{formatted_cost}")
   end
 
+  def sufficient_funds_for_sell_when_update
+    # we would want to execute this method only when we are just updating the shares
+    # we are not marking the order as completed
+
```
pollygee, line 73: We don't want to allow any orders to be updated that aren't pending, so we should add a guard clause here that bails unless the status is pending so we bail out before doing any other logic.

Or we could add a separate validation which might be preferable something like:

validate :order_is_pending, on: :update

 ... 

def order_is_pending
  errors.add(:base, "Cannot update non-pending orders") unless pending?
end

app/models/order.rb
```diff
@@ -51,6 +67,28 @@ def sufficient_funds_for_buy
     errors.add(:shares, "Insufficient funds. You have #{formatted_balance} but need #{formatted_cost}")
   end
 
+  def sufficient_funds_for_sell_when_update
+    # we would want to execute this method only when we are just updating the shares
```
pollygee, line 71: We don't want to leave these comments in, can you remove them please?
Let's aim to name the methods and create tests in such a way that we are document what's happening.

test/models/order_test.rb
```diff
@@ -169,4 +169,83 @@ class OrderTest < ActiveSupport::TestCase
...
+
+    assert_difference "PortfolioTransaction.count", 0 do
+      order.update!(shares: 3)
+    end
+
+    transaction = PortfolioTransaction.last
+    assert_equal user.portfolio, transaction.portfolio
+    assert_equal 3_000, transaction.amount_cents
+    assert_equal "credit", transaction.transaction_type
+  end
```
pollygee, line 250: I think we're missing the condition of:
Test for preventing overselling shares.  Could you add that one please?  I think this one would fail with the missing method definition but would pass once that is fixed which should keep us covered.

test "update order prevents overselling shares when updating sell order" do
  user = create(:student)
  stock = create(:stock, price_cents: 1_000)
  create(:portfolio_stock, portfolio: user.portfolio, stock: stock, shares: 5, purchase_price_cents: 1_000)

  order = build(:order, user:, stock:, shares: 2, transaction_type: "sell")
  order.save!
  
  assert_not order.valid?
  assert_includes order.errors[:shares], "Cannot sell more shares than you own"
end

pollygee (changes requested):
Thanks for the PR and for finding some issues that need to be fixed. 

In general it's helpful if we create more issues when new things are found so we can be on the same page about the fix.  And it's better if we break out the issue rather than fixing multiple on the same PR.  Then we can make sure you have the context on each PR and we can tackle smaller pieces together.  

I have added a few comments. Let me know what you think, and if anything isn't clear in what I wrote.

app/views/orders/new.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="ps-4">
-  <%= render "form", order: @order, shares_owned: @shares_owned %>
+  <%= render "form", order: @order, transaction_type: @order.transaction_type, shares_owned: @shares_owned %>
```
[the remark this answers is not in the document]
suraj-mandal: I could not understand this one. This form is specifically the edit form for editing buy/sell orders.

PR #435, opened 2025-08-30
Seed data is broken

davidpgero:
When I tried to set up the application on my local machine, I ran into two problems. 

1) Missing action for `Order` model
2) Not enough money for the buyer to purchase the first seed data, AAPL.


pollygee (approved):
This is super helpful thanks for fixing this!

Issue #424, opened 2025-08-25
Stocks source of truth


pollygee:
# Manage Stocks Through Database Instead of Hardcoded Values

## Description
Currently, stocks are partially hardcoded in the application with a SYMBOLS constant in the Stock model. There's also an admin interface for managing stocks, but we need to:

1. Add a complete set of known stocks to the seeds for both development and production environments (this list is hardcoded list in `/models/stocks.rb`
2. Remove the hardcoded list of stocks and use the database as the single source of truth

## Current Implementation
- Stocks have a hardcoded SYMBOLS constant in `app/models/stock.rb`
- Stocks are displayed in `app/views/stocks/index.html.erb` by fetching all stocks from the database
- There's an admin interface for managing stocks in `app/controllers/admin/stocks_controller.rb`
- Stocks are seeded in `db/seeds/partials/stocks.rb` with basic information
- The production seeds file (`db/seeds/production.rb`) doesn't currently seed stocks

## Requirements
1. Remove the hardcoded SYMBOLS constant from the Stock model
2. Add a complete set of known stocks to both development and production seeds
3. Update the application to use only database-stored stocks as the source of truth
4. Ensure that both development and production environments have a consistent set of seeded stocks

## Expected Outcomes
- Stocks are fully managed through the database rather than hardcoded values
- Admins can manage stocks through the existing admin interface
- The application uses only database-stored stocks as the source of truth
- Both development and production environments have seeded stocks

## Implementation Considerations
- Update seeds to include a comprehensive list of stocks for both environments
- Ensure that the admin interface continues to work for stock management
- Update any references to the hardcoded list to use database records instead

## Testing
- Verify that stocks are properly seeded in both environments
- Confirm that the admin interface works for managing stocks
- Test that the seeded stocks appear correctly in the UI
- Ensure that all stock-related functionality works as expected

## Acceptance Criteria
- [ ] Hardcoded stock list is removed from the codebase
- [ ] Stocks are managed entirely through the database
- [ ] Both development and production environments have seeded stocks
- [ ] Admin interface continues to work for stock management
- [ ] Adequate test coverage added

davidpgero:
@pollygee , I'm happy to take this ticket if it's still open.

pollygee:
Great thanks!

PR #447, opened 2025-09-01
#424 stock source of truth from the database

davidpgero:
It resolves the https://github.com/rubyforgood/stocks-in-the-future/issues/424

- Adding seed ticker
- Use the job to update the price
- Use the job to update the attributes

image

### DIFF
--- a/app/services/purchase_stock.rb
+++ b/app/services/purchase_stock.rb
@@ -16,7 +16,7 @@ class PurchaseStock
     return unless order.pending?
 
     ActiveRecord::Base.transaction do
-      create_withdrawal_transaction
+      create_portfolio_transaction
       create_portfolio_stock
       update_order_status
     end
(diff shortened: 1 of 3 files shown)


db/seeds/partials/stocks.rb
```diff
@@ -1,198 +1,10 @@
...
 
-# finding an existing stock or create using only company_name for now
-stocks.each do |stock_data|
-  stock = Stock.find_or_create_by(company_name: stock_data[:company_name])
-  stock.update!(stock_data)
+SEED_TICKERS.each do |ticker|
+  Stock.find_or_create_by(ticker: ticker)
 end
+
+StockPricesUpdateJob.perform_now
```
pollygee, line 9: We have a limited number of API calls on the free tier, so we're going to have to leave this as is I think.  Also I think it makes sense to allow people to work on the app and have seed data without having to create an API key.  So for those reasons can you revert the changes in this file please?

pollygee (changes requested):
This looks great thanks so much.
I just have a comment requesting to revert a change, but other than that it lgtm!  
When you make the update we'll change this to an approval.  Thanks again.

pollygee (approved):
Thanks for the updates!

PR #450, opened 2025-09-02
403 adding portfolio positions for aggragates view

davidpgero:
It resolves https://github.com/rubyforgood/stocks-in-the-future/issues/403

- New value object, PortfolioPostion
- Use weighted avg

## portfolio
image

## trading floor
image


davidpgero:
the issue is already solved. https://github.com/rubyforgood/stocks-in-the-future/issues/403

PR #454, opened 2025-09-04
Added status badges to the README.md file

jhsu802701:
This provides makes the current status of continuous integration, RuboCop, and Dependabot clearly visible.  Because the status badges are not in the README.md file, checking the results of these builds requires clicking on the "Actions" tab and hunting around.


pollygee (approved):
This is great thanks so much.

PR #457, opened 2025-09-04
Removed puts commands from test suite

jhsu802701:
These puts commands in the test suite clutter up the screen output.  This pull request removes puts commands from test/models/sti_test.rb.


pollygee (approved):
Great thanks!

PR #476, opened 2025-09-05
Buy or sell order safe for invalid number or shares

davidpgero:
- Fixed `nil can't be coerced into Integer` exceptions when submitting orders without shares
- Fixed `undefined method '>' for nil` exceptions on sell orders with nil shares
- Handle validation + HTML5 validation too

## Buy order without valid shares

From:

image

To:

image

## Sell order without valid shares

From:
image

To:
image

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -42,7 +42,13 @@ class Order < ApplicationRecord
 
   private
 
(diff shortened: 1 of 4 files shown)


pollygee (approved):
Looks great thanks!

Issue #462, opened 2025-09-04
Grade level on classroom normalized


pollygee:
Grade is a text field and we will have all of the following:
- Sixth Grade
- Sixth
- sixth grade
- 6th grade

Change it to be a numeric field so we can keep things consistent.

PR #506, opened 2025-09-08
#462 classroom grade is numeric

davidpgero:
it resolves https://github.com/rubyforgood/stocks-in-the-future/issues/462

- Add new grade_numeric :integer field
- use this new grade_numeric field first
- then we can remove the grade string field and rename

Screenshot 2025-09-08 at 15 22 51

Next steps are:
1. Migrate existing grade data to grade_numeric
2. Update forms/views to use grade_numeric
3. Remove old grade field and rename grade_numeric to grade


davidpgero:
> Did you decide not to merge this

Yes, I'm recreating it. 

I'll share the new PR soon, as we don't need that safe operation without a production environment.

PR #517, opened 2025-09-09
Corrected the class that was flagged as duplicated and overwritten

jhsu802701:
The test suite produces the following warning message:

`/home/runner/work/stocks-in-the-future/stocks-in-the-future/app/views/classrooms/_classroom.html.erb:25: warning: key :class is duplicated and overwritten on line 26`

The change in this pull request addresses this warning by consolidating the two class specifications into one.


dpowers1 (approved):
Love it. Thanks for the cleanup!

PR #521, opened 2025-09-09
Upgrade TailwindCSS from 3.3.2 to 4.3.0

### LINKED ISSUE #452: Tailwinds Upgrade
We need to upgrade tailwinds so we can use fancy thigs like dropdowns without having to use react 🎉 
More info [here](https://tailwindcss.com/blog/vanilla-js-support-for-tailwind-plus)

We are currently on 3.4 and we need to upgrade to at least 4.1 to get the functionality we're after.

AbuMareBear:
Resolves #452

### DIFF
--- a/app/dashboards/student_dashboard.rb
+++ b/app/dashboards/student_dashboard.rb
@@ -67,8 +67,7 @@ class StudentDashboard < Administrate::BaseDashboard
 
   # Overwrite this method to customize how students are displayed
   # across all pages of the admin dashboard.
-  #
-  # def display_resource(student)
-  #   "Student ##{student.id}"
-  # end
+  def display_resource(student)
+    student.username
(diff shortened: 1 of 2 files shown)


dpowers1 (changes requested):
Wow great work! Thanks for taking this on. I only saw one issue that I consider blocking.
The issue is the cards seem like they are all shadow? I image is on top of everything in main as well so dont worry about that
Screenshot 2025-09-09 at 7 41 16 PM

AbuMareBear:
Thank you @dpowers1 and @pollygee for taking the time to look. I'll address the merge conflicts and check the boxes this morning.

AbuMareBear:
@dpowers1 I'm not seeing the issue you're describing.

image

Issue #352, opened 2025-07-04
Add ability to bulk add students


pollygee:
We want to be able to bulk add students on the student index admin page
Somewhere at the top something like the screenshot below (use the `.tw-btn-secondary` css class)

We only need uesername and classroom and lets use ID for the classroom, I will train the admins to use the IDs

Don't worry about error handling right now, that will be a separate issue.

![Image](https://github.com/user-attachments/assets/5afdc405-1453-4bb3-8744-95e044cd552a)

Issue #311, opened 2025-07-02
Password generation improvements


pollygee:

pollygee:
Currently we have a short list of words that is visible in GH to create the student passwords this isn't very secure.  Having a longer list that isn't stored in GH would help us make this more secure.  There are other things we could do as well.  We need the password to be easy to read out/share/remember. So that is a big limitation for us.

@dpowers1 what are your thoughts on this?

dpowers1:
We could use something like https://memorablepasswordgenerator.com/ to generate them on the fly (this was the best tool I could find) 

We could also build the word list out and randomly pick a combo of 3 words

pollygee:
Because the kids are young I think we need to try to keep it simple.  A very long list of words to choose from would be an improvement.  And what I have in there now is combining that with a number instead of 2 words.  We can make a final call next week in person.

PR #522, opened 2025-09-10
352 import students from CSV file

davidpgero:
it resolves https://github.com/rubyforgood/stocks-in-the-future/issues/352

image

### DIFF
--- a/app/services/friendly_password_generator.rb
+++ b/app/services/friendly_password_generator.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class FriendlyPasswordGenerator
+  ADJECTIVES = %w[happy quick brave smart cool nice warm bright fresh clean].freeze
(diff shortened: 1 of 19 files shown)


app/services/friendly_password_generator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class FriendlyPasswordGenerator
```
davidpgero, line 3: I'm not sure we need it. Maybe we can remove it. 

We just need a password to create a new student account. 

Maybe it will come from the CSV or be generated by the system and send out an email... There are too many questions at the moment. 

What do you think?

davidpgero: It can be related to that, too:

https://github.com/rubyforgood/stocks-in-the-future/issues/311

Issue #560, opened 2025-09-12
Portfolio table Change and Total Return columns


pollygee:
Update the calculations for columns `Change` and `Total Return` 

Change ----->  how much I have earned or lost from when I purchased total for all the stocks I've purchased in this category

Total Return ------>  how much it's worth if you sell it right now
(Qty * Last Price)

PR #571, opened 2025-09-13
Fix Portfolio table Change and Total Return columns #560

ErinClaudio:
Fix Portfolio table Change and Total Return columns #560
Problem:
  Change column showed $0.00 due to broken calculation: current_price - current_price = 0

  Solution:
  - Fixed calculation: (current_price - purchase_price) × shares
  - Added color coding (green gains, red losses)
  - Handle cents/dollars format mismatch

  Changes:
  - Added change_amount/total_return_amount methods to PortfolioStock
  - Updated portfolio view with proper calculations and styling
  - Added comprehensive test coverage

  Result:
  ✅ All tests pass, accurate profit/loss display with proper formatting
  
  ```
Note: current seed data will show $0.00 in Change column 

  To test the fix functionality:
  1. Update a stock price: bin/rails runner "Stock.find_by(ticker: 'BAC').update(price_cents: 5500)"
  2. View Mike's portfolio (mike@example.com / password)
  3. Verify: BAC now shows green positive change, other stocks remain $0.00

```

I am happy to change the seed data so that these numbers show a change. What you see below is a local change 

Screenshot 2025-09-16 at 12 17 51 PM

### DIFF
--- a/app/models/portfolio_stock.rb
+++ b/app/models/portfolio_stock.rb
@@ -6,6 +6,8 @@ class PortfolioStock < ApplicationRecord
 
   def change_amount
     current_price = stock.current_price
+    # TODO: this is a smell....why do we need to check if we have cents or dollars
+    # we must fix this.
     # Handle cents/dollars format mismatch in purchase_price
     normalized_purchase_price = purchase_price > 1000 ? purchase_price / 100.0 : purchase_price
     (current_price - normalized_purchase_price) * shares
--- a/app/models/stock.rb
+++ b/app/models/stock.rb
@@ -15,6 +15,7 @@ class Stock < ApplicationRecord
 
   def yesterday_price
     return current_price if yesterday_price_cents.nil?
+
     yesterday_price_cents.to_f / 100
   end
 end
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by pollygee** [comment]
@ErinClaudio can you change this to a draft PR till you're ready for review please?

File: app/models/portfolio_stock.rb [hunk 1]
```diff
@@ -4,6 +4,16 @@ class PortfolioStock < ApplicationRecord
   belongs_to :portfolio
   belongs_to :stock
 
+  def change_amount
+    current_price = stock.current_price
+    normalized_purchase_price = purchase_price > 1000 ? purchase_price / 100.0 : purchase_price
```
pollygee, line 9: I'm confused here.  `purchase_price` is stored in the db and is a decimal right?

pollygee (approved):
Thanks for wrapping this up Erin!  

I'm not blocking the merge but I am confused why we're doing the division check?  I can't figure out what's happening there.  If we need it this way maybe we need a comment?

app/models/portfolio_stock.rb
ErinClaudio: Screenshot 2025-09-26 at 9 33 16 AM

Issue #575, opened 2025-09-13
We need to store yesterday's stock price


pollygee:
Add a column to Stocks `yesterdays_price`
When we run the job to grab the new stock price, before making the API call store the current price as yesterday's price.

PR #597, opened 2025-09-15
Store yesterday's price

davidpgero:
It resolves https://github.com/rubyforgood/stocks-in-the-future/issues/575

It adds yesterday's price to the stock
We store yesterday's price when we updating the price

### DIFF
--- a/app/jobs/stock_prices_update_job.rb
+++ b/app/jobs/stock_prices_update_job.rb
@@ -6,44 +6,49 @@ require "json"
 class StockPricesUpdateJob < ApplicationJob
   queue_as :default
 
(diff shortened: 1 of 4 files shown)


app/jobs/stock_prices_update_job.rb
```diff
@@ -6,44 +6,50 @@
 class StockPricesUpdateJob < ApplicationJob
   queue_as :default
 
-  # rubocop:disable Metrics/AbcSize
   def perform(...)
```
pollygee, line 9: Thanks for refactoring so we can get rid of this rubocop disable 🥳

pollygee (approved):
Looks great. Thanks!

PR #602, opened 2025-09-16
Fix soft delete issues with discarded students

### LINKED ISSUE #596: Use soft delete when deleting students from Student Show page and admin clasroom show page
When we delete a student from the classroom show page both in the app and in the admin side, the student is discarded properly. However, it still shows up there so it appears as if it hasn't been removed. I think the expected behavior would be that the students would only appear on this page if they have not been deleted. Additionally you can discard a student who is already discarded which causes an error.

I don't think we need to worry about being able to undo this action - an admin can do that if it's necessary but I do think we should filter the classroom show page.

This is a follow up to #550

ersync:
## Problem
After implementing soft delete, three issues remained:
1. Discarded students still showed on classroom pages
2. Clicking "Delete" on already-deleted students crashed the app
3. Classroom stats counted discarded students

## Changes
- `StudentsController#set_student` uses `.kept.find()` with error handling
- `ClassroomsController#show` and stats use `.kept` scope  
- `Classroom#students` association scoped to exclude discarded records
- Added `students.not_found` translation key
- Added unit and integration tests for discarded student scenarios
```ruby
# Model test: 
test "students association excludes discarded students"

# Controller test: 
test "teacher is redirected when accessing discarded student"
```
## Result
- Discarded students disappear from classroom pages immediately
- No more crashes when trying to delete already-deleted students
- Accurate classroom statistics
- User-friendly "Student not found" messages

Closes #596

### DIFF
--- a/app/controllers/classrooms_controller.rb
+++ b/app/controllers/classrooms_controller.rb
@@ -12,7 +12,7 @@ class ClassroomsController < ApplicationController
   end
 
   def show
-    @students = @classroom.users.students.includes(:portfolio, :orders)
+    @students = @classroom.users.students.kept.includes(:portfolio, :orders)
     @can_manage_students = current_user.teacher_or_admin?
     @classroom_stats = calculate_classroom_stats if @can_manage_students
   end
@@ -91,7 +91,7 @@ class ClassroomsController < ApplicationController
   def calculate_classroom_stats
     return {} unless @classroom
 
-    students = @classroom.users.students
+    students = @classroom.users.students.kept
     {
       total_students: students.count,
       active_students: students.joins(:orders).distinct.count,
--- a/app/models/classroom.rb
+++ b/app/models/classroom.rb
@@ -12,7 +12,7 @@ class Classroom < ApplicationRecord
   has_many :users, dependent: :nullify
   has_many :teacher_classrooms, dependent: :destroy
   has_many :teachers, through: :teacher_classrooms
-  has_many :students, class_name: "Student", dependent: :nullify
+  has_many :students, -> { kept }, class_name: "Student", inverse_of: :classroom, dependent: :nullify
   has_many :grade_books, dependent: :nullify
 
   validates :name, presence: true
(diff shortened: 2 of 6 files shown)


test/models/classroom_test.rb
```diff
@@ -91,4 +91,15 @@ class ClassroomTest < ActiveSupport::TestCase
...
+
+  test "students association excludes discarded students" do
+    classroom = create(:classroom)
+    kept_student = create(:student, classroom: classroom)
+    discarded_student = create(:student, classroom: classroom)
+    discarded_student.discard
+
+    assert_includes classroom.students, kept_student
+    assert_not_includes classroom.students, discarded_student
+    assert_equal [kept_student.id], classroom.students.pluck(:id)
```
davidpgero, line 103: Detailed test, thanks! 🙇

app/models/classroom.rb
```diff
@@ -12,7 +12,7 @@ class Classroom < ApplicationRecord
   has_many :users, dependent: :nullify
   has_many :teacher_classrooms, dependent: :destroy
   has_many :teachers, through: :teacher_classrooms
-  has_many :students, class_name: "Student", dependent: :nullify
+  has_many :students, -> { kept }, class_name: "Student", inverse_of: :classroom, dependent: :nullify
```
davidpgero, line 15: i like the naming, `kept` here, because "students kept in the classroom" 🙇

app/controllers/classrooms_controller.rb
```diff
@@ -91,7 +91,7 @@ def assign_school_year_to_classroom
   def calculate_classroom_stats
     return {} unless @classroom
 
-    students = @classroom.users.students
+    students = @classroom.users.students.kept
```
davidpgero, line 94: It's good that you fix this one too 🚀

pollygee:
Thanks for checking @ersync!  @davidpgero has been around for a bit and is reliable!  I'm grateful for the help with reviews, so yes you're good to go!

Issue #588, opened 2025-09-14
Create a job to run monthly  to store the student's portfolio value


pollygee:
We need to record the montly value for each students portfolio so eventally we can make graphs

PR #630, opened 2025-09-23
588 monthly portfolio snapshots jobs

davidpgero:
it resolves https://github.com/rubyforgood/stocks-in-the-future/issues/588

It runs the monthly job at the end of month, but not trading calender day.


pollygee (approved):
Looks great thanks!

PR #635, opened 2025-09-24
Improve application policy

davidpgero:
no functional change, only refactoring and test coverage a bit

### DIFF
--- a/test/policies/application_policy_test.rb
+++ b/test/policies/application_policy_test.rb
@@ -3,31 +3,127 @@
 require "test_helper"
 
 class ApplicationPolicyTest < ActiveSupport::TestCase
-  test "teacher_or_admin_required?" do
-    admin = create(:admin)
-    teacher = create(:teacher)
-    student = create(:student)
-    the_nil = nil
-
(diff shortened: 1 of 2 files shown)


test/policies/application_policy_test.rb
```diff
@@ -3,31 +3,127 @@
 require "test_helper"
 
 class ApplicationPolicyTest < ActiveSupport::TestCase
+  def setup
+    @admin = create(:admin)
+    @teacher = create(:teacher)
+    @student = create(:student)
+    @record = :some_record
+    @the_nil = nil
+  end
```
AbuMareBear, line 12: In my experience, having setup methods like these become a testing anti-pattern in meaningful test suites. In my opinion, DRY should be abandoned for testing and each test should create the records it needs for that test. Each test should be self-contained and readable from top to bottom.

pollygee: I understand where you're coming from @AbuMareBear  This could be overkill and could potentially make it more brittle, but in this case I think it's unlikely for that to be a problem.  I think it's prob fine here.

PR #638, opened 2025-09-25
Added a GitHub workflow for running the tests with SimpleCov

jhsu802701:


davidpgero:
Thank you for sharing your opinions! 🙇 

@jhsu802701 , I hope you won't mind, but I'd like to close this PR with this comment. 

Even if I like it and it works well, I also don't think it's beneficial to the project at this time. 

We can reopen it if the situation changes.

I'm really glad that you brought up the topic and created a PR about the solution. 

Thank you!

Issue #641, opened 2025-09-27
Create Quarters when SchoolYear is created


pollygee:
Upon creation of SchoolYear we want to create the associated quarters for that school year

Add a callback to create the four quarters for a school year when the SchoolYear is created.

pollygee:
@davidpgero let me know what you think about this approach for adding quarters. 
I don't think we want admins to have to add them.  Quarters are just periods of time, and they are consistent-there's always 4 for every schoolyear I think it makes sense to automate adding them.  What do you think?

davidpgero:
Good idea, @pollygee ! I agree on that. It's better to automate instead of leaving them to the admin. 🙇

PR #656, opened 2025-09-30
Implement SchoolYearCreationService to create school years and quarters

AlexeyKasianenko:
#641


pollygee:
resolves #641 


Thanks for the PR @AlexeyKasianenko 
it's important to use the word "resolves" before the issue number so that the issue gets closed when the PR is merged.  Thank you!  I've done it with this comment but for future it would be great if you can do it in the PR description.  It's also helpful if you fill out the description with the changes you made to address the issue. Thank you!

PR #658, opened 2025-09-30
Fix solid queue config

pollygee:
Fix Proc file 
and solid queue can't read "last day of the month" so needed to update that job to use cron format

Plus I added a couple of tests


dpowers1 (approved):
:shipit:

PR #665, opened 2025-10-01
[no-ticket] fix grade books show typo -- malformed html

jwshinx:
Small polish to html. There is an errand `>` div closing tag that this addresses. Its trivial enough that I assumed wouldnt need a formal issue. Let me know if I do.

Additionally adjusted indentation for readability.


pollygee (approved):
Thanks @jwshinx

Issue #669, opened 2025-10-01
Add REASON enums


dpowers1:
Make an enum with these field

MATH_EARNINGS: "Earnings from math"
READING_EARNINGS: "Earnings from reading"
ATTENDANCE_EARNINGS: "Earnings from attendance"
TRANSACTION_FEE: "Transaction fee"
AWARDS: "Award"

PR #675, opened 2025-10-01
Add reasons enum

dpowers1:
Closes: #669

### DIFF
--- a/app/models/portfolio_transaction.rb
+++ b/app/models/portfolio_transaction.rb
@@ -11,8 +11,9 @@ class PortfolioTransaction < ApplicationRecord
     reading_earnings: "Earnings from Reading",
     attendance_earnings: "Earnings from Attendance",
     grade_earnings: "Earnings from Grades", # TODO: Remove this
-    transaction_fee: "Transaction Fee",
-    awards: "Award"
+    transaction_fees: "Transaction Fees",
+    awards: "Award",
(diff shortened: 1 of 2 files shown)


app/models/portfolio_transaction.rb
```diff
@@ -6,6 +6,15 @@ class PortfolioTransaction < ApplicationRecord
   # credit/debit is for stock transactions ie buy/sell stocks
   enum :transaction_type, { deposit: 0, withdrawal: 1, credit: 2, debit: 3, fee: 4 }
 
+  REASONS = {
+    math_earnings: "Earnings from Math",
+    reading_earnings: "Earnings from Reading",
+    attendance_earnings: "Earnings from Attendance",
+    grade_earnings: "Earnings from Grades", # TODO: Remove this
+    transaction_fee: "Transaction Fee",
```
pollygee, line 14: should this be plural like the others?

app/models/portfolio_transaction.rb
```diff
@@ -6,6 +6,15 @@ class PortfolioTransaction < ApplicationRecord
...
+    awards: "Award"
```
pollygee, line 15: Do we need one more that is 
administrative_adjustments or something like that?  for when there's been an error and we need to add or subtract from an account to "make it right"?

pollygee (approved):
Looks good.
I had nonblocking thoughts/questions.

Issue #660, opened 2025-09-30
Look at improvement from the previous year when finalizing a grade for fund distribution


dpowers1:
app/services/distribute_earnings.rb:33 looks at previous quarters in the gradebook. 

If we are in Q1 of a year, then look at the previous year Q4 for the previous_quarter information

Also ensure we update the documentation at docs/gradebook-earnings.md

jwshinx:
## The Problem

The current `previous` and `next` methods **only work within the same school year**. They look for quarters with `number - 1` or `number + 1` within the same `school_year`.

**What happens now:**
- If you're in Q1 and call `previous`, it looks for Q0 (doesn't exist) → returns `nil`
- If you're in Q4 and call `next`, it looks for Q5 (doesn't exist) → returns `nil`

**What you need:**
- Q1 2022-2023 `previous` should return Q4 2021-2022
- Q4 2021-2022 `next` should return Q1 2022-2023

**The core issue:** There's no mechanism to navigate across different school years, even though quarters naturally flow from one academic year to the next in a continuous sequence.

This is exactly why you need either:
- **Option 1**: Global sequence numbers for simple arithmetic navigation
- **Option 2**: Smart logic to find the previous/next year and navigate to the appropriate quarter
- **Option 3**: Virtual navigation

The current implementation has a "year boundary wall" that prevents natural quarter progression across academic years.

## Architectural Approaches

### Option 1: Global Quarter Sequence
Add a global sequence number to quarters that spans across all years.

**Implementation:**
```
# Migration to add global_sequence to quarters
add_column :quarters, :global_sequence, :integer
add_index :quarters, :global_sequence, unique: true

# Example data:
# 2021-2022 Q1: global_sequence = 1
# 2021-2022 Q2: global_sequence = 2
# 2021-2022 Q3: global_sequence = 3
# 2021-2022 Q4: global_sequence = 4
# 2022-2023 Q1: global_sequence = 5
# etc.
```

**Pros:**
- Simple navigation: `quarter.global_sequence - 1` for previous quarter
- Easy to implement and understand
- Efficient queries
- Natural ordering for reports across years

**Cons:**
- Requires data migration to populate existing records
- Need to maintain sequence integrity when adding new quarters
### Option 2: Year Ordering with Smart Navigation
Add an ordering system to years and implement navigation logic in the model.

**Implementation:**
```ruby
# Migration to add start_year to years table
add_column :years, :start_year, :integer
[...]

PR #677, opened 2025-10-01
Issue-660: smarter quarters to allow next and previous navigation across years

jwshinx:
## The Problem

The current `previous` and `next` methods **only work within the same school year**. They look for quarters with `number - 1` or `number + 1` within the same `school_year`.

**What happens now:**
- If you're in Q1 and call `previous`, it looks for Q0 (doesn't exist) → returns `nil`
- If you're in Q4 and call `next`, it looks for Q5 (doesn't exist) → returns `nil`

**What we need:**
- Q1 2022-2023 `previous` should return Q4 2021-2022
- Q4 2021-2022 `next` should return Q1 2022-2023

**The core issue:** There's no mechanism to navigate across different school years, even though quarters naturally flow from one school year to the next in a continuous sequence.

## Considerations
- Im lukewarm about using keywords like `next`. So I prefer `next_thing`. OTOH I also understand the value of consistency. I know `year.previous_year` looks strange. I can be easily convinced!

## Reference
Find details in [Issue 660](https://github.com/rubyforgood/stocks-in-the-future/issues/660)

### DIFF
--- a/app/models/quarter.rb
+++ b/app/models/quarter.rb
@@ -11,10 +11,48 @@ class Quarter < ApplicationRecord
   scope :ordered, -> { order(:number) }
 
   def next
+    return first_quarter_of_next_year if number == 4
+
     school_year.quarters.ordered.find_by(number: number + 1)
   end
 
   def previous
+    return last_quarter_of_previous_year if number == 1
+
     school_year.quarters.ordered.find_by(number: number - 1)
+  end
+
+  private
+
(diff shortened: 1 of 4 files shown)


jwshinx:
@pollygee @dpowers1 this is just ruby work, but i didnt end-to-end test the ux. it felt like a bit of effort to set up my local with proper data -- and hit the `finalize` endpoint. just a headsup, as you review

PR #680, opened 2025-10-01
School name should be required

americac:


pollygee (approved):
Great thanks for so much test coverage!

PR #700, opened 2025-10-03
Sagarp2901 fix/619 navbar pundit update

pollygee:


pollygee:
No worries at all.  I was just concerned because we have so much work happening right now we keep handing you more merge conflicts!
I'm not sure what that error is but if you run into it again just let us know we can try to help!

Issue #698, opened 2025-10-03
Update school admin page


pollygee:
We have two tables on the show school admin page and we want only one, there's redundant info 

- [ ] remove the second table labeled Years
- [ ] remove school column 
- [ ] have created at just be a date, no time needed
- [ ] remove created  and updated at from the bottom of the page

PR #705, opened 2025-10-04
Fixes Issue #698

suraj-mandal:
This PR fixes issue: [#698 ](https://github.com/rubyforgood/stocks-in-the-future/issues/698).

As of now the following changes have been implemented:

- created_at is being shown as `date` rather than `datetime`
- the `school year` dashboard is not been shown here.
- removed the `school year` route from administrate.
- school dashboard shows `name`, `years` and `created at`.

Screenshot From 2025-10-04 11-56-54

Do let me know if further changes are required.


pollygee:
@suraj-mandal I don't think there is anything on here about removing a route.  We don't want to remove any routes at all.  
We just don't want that second table to show that was auto-generated by administrate.

suraj-mandal:
Understood.. fixing it

suraj-mandal:
Sure, working on this now!

app/views/admin/application/_navigation.html.erb
```diff
@@ -0,0 +1,21 @@
...
+By default, the navigation contains navigation links
+for all resources in the admin dashboard,
+as defined by the routes in the `admin/` namespace
+%>
+
+<nav class="navigation">
+  <%= link_to(t("administrate.navigation.back_to_app"), root_url, class: "button button--alt button--nav") if defined?(root_url) %>
+
+  <% Administrate::Namespace.new(namespace).resources_with_index_route.each do |resource| %>
+    <% next if resource.to_sym == :school_years %>
```
pollygee, line 14: @suraj-mandal I don't think this partial was in the previous PR.

I think there's a misunderstanding here again.  Sorry about that.  The ticket is just asking for the school_years to be removed from the admin school show page because the second table taht was showing up didn't make any sense.  

This partial is removing schoo_years from the navigation and we don't want that.  We are keeping school years only just not showing it on that page.

Show_School__2_-_Stocks_In_The_Future

The table in the box is redundant because we already have school and years in the table above.  That is the only thing we're removing.  Just this table on this page.

Sorry for the back and forth on this one.  I don't think I saw this additional partial on the PR the previous time you had it up.  You can remove this file.  And the test.  Actually the test is great coverage for our admin navigation you can just update it to the affirmative for  link for school years.

suraj-mandal: Sure.. I got it.. doing the fix by tonight.. thanks a lot for the clarification

suraj-mandal: As of now, I have removed the custom navigation bar as you have suggested. Let me know in case there is anything extra required. Thanks!

pollygee (approved):
Thanks for the changes

PR #709, opened 2025-10-04
Add color coding to Total Return column in portfolio table

### LINKED ISSUE #684: Portfolio table Rotal Return colmn should show as red or green
Image

Positive quantities for Total Return should show green, negative in red just like the change column

ersync:
Added red/green color styling to the Total Return column to match the Change column. Positive returns show in green with a + prefix, negative returns show in red.

Fixes #684


davidpgero (approved):
It looks good! Thank you! 🙇

Issue #710, opened 2025-10-04
Trade from Stock is broken


davidpgero:
Trading flow is broken if I start from the Stock view.

### Trading floor

Log in as `mike` with `password`. 

Then go to the trading floor (http://localhost:3000/stocks)

You can buy/sell shares 

<img width="800" height="214" alt="Image" src="https://github.com/user-attachments/assets/e0192090-cab4-4304-beeb-6312262ff5be" />

### Stock view

But from the stock view, it's broken. 

For instance, http://localhost:3000/stocks/1

Click the trade button:

<img width="800" height="163" alt="Image" src="https://github.com/user-attachments/assets/5c530def-4bc8-415b-96b7-d55ee3eddf05" />

I can only sell, not buy

<img width="800" height="551" alt="Image" src="https://github.com/user-attachments/assets/04335b58-e62d-430c-96c9-81bfd9c5202f" />

And then I end up on an error where the action type is missing. 

<img width="800" height="607" alt="Image" src="https://github.com/user-attachments/assets/05d0951a-15bd-4843-a4e8-00f9a92c2e9a" />

davidpgero:
The main question is how we want to handle when a student starts trading from the Stock's view.

PR #715, opened 2025-10-06
710 buy/sell buttons + remove trade button

davidpgero:
It resolves https://github.com/rubyforgood/stocks-in-the-future/issues/710

The trade button does not function in multiple places due to the missing modal. 

So, as I first step, I create a buy/sell button, then we can make a trade button where you can select buy/sell

- one buy button, multiple reuse
- one sell button, multiple reuse
- Remove the unified trade button until we know how we want to implement it

### my portfolio
image

### trading floor
image

### stock view

image

## Next step

- [ ] Create a ticket about the unified trade button design? After clicking on 'Trade', you can then select 'Buy/Sell'.


pollygee:
🤔 
The trade button was supposed to take you to the trading floor
It's also not supposed to show up for teachers and admins because they can't trade anything
I just had a discussion with the designer today and she thinks the buttons should still show but should be deactivated but I wasn't convinced yet 😆 

I think it's prob fine to have buy and sell buttons in all these places but we should check with her.  I'll try to connect with her tomorrow

PR #727, opened 2025-10-08
Add missing staging.rb seeds file so we can seed in staging

pollygee:
### Problem
Heroku deployments to staging environment were failing when running `heroku rake db:seed` because we didn't have a staging file

### Changes
Added: db/seeds/staging.rb - loads development seeds for staging environment
We want all the sample data we have for development in staging so we can play around with it


maebeale (approved):
TIL about staging.rb!

davidpgero (approved):

PR #733, opened 2025-10-10
663 Fix archived stock behavior for students

### LINKED ISSUE #663: Check archived stocks
Check that the following are true and update anywhere that archived stocks are not working correctly:

Students **can not** buy archived stocks (buy button should be deactivated on archived stocks)
Students **can** sell archived stocks (sell button is available)
Students **can** hold archived stocks (if they have holdings it should show on their portfolio page)
Students **can** cash out their portfolio with archived stocks holdings 
Monthly job should be considering any holdings that students have including the archived stocks

We want to see archived stocks on the stocks index page but they should show at the bottom of the page.  So we should first iterate through active stocks then iterate through inactive stocks

hagiya0121:
Resolves #663  

### Description
- Students can no longer buy archived stocks (Buy button hidden)  
- Students can sell archived stocks they already own (Sell button available)  
- Active and archived stocks are displayed in separate tables  
- All students can view archived stocks   

### Screenshots
image

### DIFF
--- a/app/policies/stock_policy.rb
+++ b/app/policies/stock_policy.rb
@@ -50,11 +50,7 @@ class StockPolicy < ApplicationPolicy
   # Scope to control which stocks are visible in listings
   class Scope < ApplicationPolicy::Scope
     def resolve
-      if user&.admin? || user&.teacher? || user&.student?
-        scope.all
-      else
-        scope.active
-      end
+      scope.all
     end
   end
 end
--- a/test/policies/stock_policy_test.rb
+++ b/test/policies/stock_policy_test.rb
@@ -109,33 +109,24 @@ class StockPolicyTest < ActiveSupport::TestCase
     end
   end
 
-  test "Scope resolves: admins and teachers see all; others see active only" do
+  test "Scope resolves: all logged-in users see all stocks" do
     admin = create(:admin)
     teacher = create(:teacher)
     student = create(:student)
 
-    active1 = create(:stock, archived: false)
-    active2 = create(:stock, archived: false)
+    active = create(:stock, archived: false)
     archived = create(:stock, archived: true)
 
     admin_scope = StockPolicy::Scope.new(admin, Stock.all).resolve
-    assert_includes admin_scope, active1
-    assert_includes admin_scope, active2
+    assert_includes admin_scope, active
     assert_includes admin_scope, archived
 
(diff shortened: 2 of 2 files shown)


app/controllers/stocks_controller.rb
```diff
@@ -6,7 +6,7 @@ class StocksController < ApplicationController
   before_action :set_portfolio
 
   def index
-    @stocks = policy_scope(Stock).includes(portfolio_stocks: :portfolio)
+    @stocks = Stock.all
```
davidpgero: I'm not sure we want to remove authorization via pundit policy. Can you help me a bit with that?

app/models/user.rb
```diff
@@ -64,6 +64,10 @@ def email_changed?
     false
   end
 
+  def holding?(stock)
+    portfolio&.portfolio_stocks&.exists?(stock_id: stock.id)
```
davidpgero: It's showing holding true even if I sell all of my stocks 🤔 

If you check the PortfolioPosition, we make sure to show only stock that has `.having("SUM(portfolio_stocks.shares) > 0")`, so if I had 2, then I sold 2, then I don't have that stock anymore.

davidpgero (commented):
Thank you for working on that! 

I have a couple of questions that I added. 

Can you take a look at them, please? 

Thank you! 🙇

app/controllers/stocks_controller.rb
hagiya0121: @davidpgero 
You’re right — it’s better to keep the Pundit policy in place.
I’ll update the StockPolicy::Scope so that students can also see archived stocks instead.
How does that sound?

pollygee:
@davidpgero can you look to see if all the requested changes have been fufilled?

app/policies/stock_policy.rb
```diff
@@ -50,7 +50,7 @@ def portfolio_present?
   # Scope to control which stocks are visible in listings
   class Scope < ApplicationPolicy::Scope
     def resolve
-      if user&.admin? || user&.teacher?
+      if user&.admin? || user&.teacher? || user&.student?
```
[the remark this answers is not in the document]
hagiya0121: @davidpgero 
You’re right — since all authenticated users can see every stock, the if statement isn’t necessary anymore.
I’ve simplified it to just scope.all for clarity.

Issue #671, opened 2025-10-01
Break out improvement bonus in services/distribute_earnings to create seperate transactions for math, reading, and attendance


dpowers1:
Rollup the different types of earnings eg.
  - math_earnings = math_grade + math_improvement

Create separate transactions for each type. one for math, reading, and attendance

pollygee:
This ticket is open to be picked up

PR #739, opened 2025-10-12
671 separate earning transactions

davidpgero:
It resolves https://github.com/rubyforgood/stocks-in-the-future/issues/671

- GradeEntry handles reading and math improvements
- DistributeEarning adding transactions with reason


pollygee:
@AbuMareBear do you have any reservations to approve this?  I don't think that the discussion happening should block this.  We can do follow up work to the discussion if need be separately.

PR #758, opened 2025-10-15
Archive classroom

### LINKED ISSUE #749: Don't allow classroom to be deleted
Currently we allow classrooms to be deleted.  After a conversation with the stakeholder, we are learning that there are reasons they will want this historical data.

For this ticket we want to remove the delete classrooms routes, and dissallow deleting a classroom.
We also need an archived flag on classroom.

Only an admin can mark a classroom as archived and we can change the delete classroom button on the admin index and show pages to be be archive.

In a followup ticket we will handle the fact that this means that students will need to be able to belong to multiple classrooms.

Barsu5489:
- Delete routes removed.
- Disallow deleting classrooms.
- Added archived flag.
- Admin-only archive functionality.
- Replaced delete with archive buttons.
- Teachers only see active (non-archived) classrooms.
- Historical data preserved through archiving instead of deletion
Fixes #749

### DIFF
--- a/app/jobs/monthly_portfolio_snapshot_job.rb
+++ b/app/jobs/monthly_portfolio_snapshot_job.rb
@@ -4,10 +4,16 @@ class MonthlyPortfolioSnapshotJob < ApplicationJob
   queue_as :default
 
   def perform(target_date = Date.current, batch_size = 1000)
+    Rails.logger.info "[RECURRING JOB] MonthlyPortfolioSnapshotJob starting at #{Time.current} (#{Time.zone})"
+    Rails.logger.info "[RECURRING JOB] Current environment: #{Rails.env}"
+    Rails.logger.info "[RECURRING JOB] Target date: #{target_date}"
+
     Portfolio.includes(:portfolio_stocks, :stocks)
              .find_in_batches(batch_size: batch_size) do |batch|
       batch.each { |portfolio| create_snapshot_for_portfolio(portfolio, target_date) }
     end
+
+    Rails.logger.info "[RECURRING JOB] MonthlyPortfolioSnapshotJob completed successfully"
   end
 
   private
--- a/app/jobs/order_execution_job.rb
+++ b/app/jobs/order_execution_job.rb
@@ -6,7 +6,8 @@ class OrderExecutionJob < ApplicationJob
   retry_on StandardError, wait: :exponentially_longer, attempts: 3
 
   def perform
-    Rails.logger.info "Starting order execution job at #{Time.current}"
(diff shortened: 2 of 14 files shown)


app/controllers/admin/classrooms_controller.rb
```diff
@@ -44,5 +44,17 @@ class ClassroomsController < Admin::ApplicationController
 
     # See https://administrate-demo.herokuapp.com/customizing_controller_actions
     # for more information
+
+    def archive
```
davidpgero: Maybe using `toggle_archive` as a naming or a separate action could be helpful here. 

In this case, we need to check whether it is already archived or not in the view layer.

What do you think?

app/models/classroom.rb
```diff
@@ -17,7 +17,18 @@ class Classroom < ApplicationRecord
...
   validates :name, presence: true
 
+  scope :active, -> { where(archived: false) }
+  scope :archived, -> { where(archived: true) }
+
   def grade_display
     grade&.ordinalize
   end
+
+  def archive!
```
davidpgero: I'm not sure it's necessary here. 

What if we just set up `archived: true` and call the save on the controller?

What do you think?

davidpgero (commented):
It's a good direction! Thank you! 

Please, check the CI. 

- It breaks a couple of tests that we need to update. 
- Also, there is no test for the new methods
- And we have an issue with the linter.

app/views/admin/classrooms/show.html.erb
```diff
@@ -0,0 +1,69 @@
...
+        method: :patch,
+        data: { turbo_confirm: "Are you sure you want to archive this classroom?" }
+      ) %>
+    <% end %>
+  </div>
+</header>
+
+<section class="main-content__body">
+  <dl>
+    <% page.attributes.each do |attribute| %>
```
davidpgero: I think it has a different API 🤔 

`page.attributes.each do |title, attributes|`

app/controllers/admin/classrooms_controller.rb
Barsu5489: `toggle_archive` makes sense because the method flips states.
I'll update to use this naming.

app/models/classroom.rb
Barsu5489: Much cleaner.

davidpgero (changes requested):
There is one test which needs to removed. Please, check the CI

https://github.com/rubyforgood/stocks-in-the-future/blob/2901b6be3f2f56a0b0f338089fc49397f44be5cd/test/system/user_manages_classrooms_test.rb#L61-L72

Can you fix that?

PR #788, opened 2025-10-21
Enforce first method argument line break

### LINKED ISSUE #746: Team Discussion: Should we enforce first method argument line breaks?
# Team Discussion: Should we enforce first method argument line breaks?

## Question for the Team
Should we standardize our multiline method call formatting by requiring the first argument to be on its own line? I'd love to hear everyone's thoughts on this!

## What I'm seeing in our codebase
I've noticed we have inconsistent formatting for multiline method calls. Here's an example I came across recently:
```ruby
current_entry = create(:grade_entry,
                       grade_book: current_grade_book,
                       user: student,
                       attendance_days: 12,
                       math_grade: "B",
                       reading_grade: "A")
```

## My preference
Personally, I prefer this style because it feels more consistent and readable to me:
```ruby
current_entry = create(
  :grade_entry,
  grade_book: current_grade_book,
  user: student,
  attendance_days: 12,
  math_grade: "B",
  reading_grade: "A"
)
```

## How we could implement this
If we decide to move forward, we could use the `Layout/FirstMethodArgumentLineBreak` RuboCop rule:

```yaml
Layout/FirstMethodArgumentLineBreak:
  Enabled: true
```

We might also want to consider these complementary rules:

```yaml
Layout/MultilineMethodCallIndentation:
  Enabled: true
  EnforcedStyle: indented

Layout/FirstArgumentIndentation:
  Enabled: true
  EnforcedStyle: consistent
```

## Questions for discussion
- Do you prefer the current mixed style or would you like consistency?
- If we standardize, which format do you prefer and why?
- Are there any concerns about making this change (impact on existing code, readability, etc.)?
- Should we auto-fix existing code or just enforce it going forward?

## My reasoning
I think the second format is cleaner because:
- The method name and arguments are visually separated
- It's easier to scan the argument list
- The closing parenthesis aligns nicely with the method call

But I'm curious what everyone else thinks! **What's your preference and why?**

This issue was written with the help of AI.

AbuMareBear:
Resolves #746

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,6 +27,9 @@ Style/Documentation:
 Style/StringLiterals:
   EnforcedStyle: double_quotes
 
+Layout/FirstMethodArgumentLineBreak:
+  Enabled: true
+
 Layout/LineLength:
   Max: 120
 
--- a/app/controllers/classrooms_controller.rb
+++ b/app/controllers/classrooms_controller.rb
@@ -90,8 +90,10 @@ class ClassroomsController < ApplicationController
       total_students: students.count,
       active_students: students.joins(:orders).distinct.count,
       total_portfolio_value: students.joins(:portfolio).sum("portfolios.current_position"),
-      recent_orders_count: Order.joins(:user).where(users: { classroom: @classroom }).where("orders.created_at > ?",
-                                                                                            1.week.ago).count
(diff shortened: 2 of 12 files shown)


davidpgero (commented):
~Can you resolve the conflict, please? Then it's good to go.~  forget to use request changes :D

PR #801, opened 2025-10-22
Fix Total Stock Buys calculation on portfolio page

### LINKED ISSUE #798: BUG: Total stock buys is not accurate
UPdate calculation on stock buys on the portfolio page to be accurate
currently it's showing 0 when a student has holdings they had to have bought them.

Image

heyjosephme:
## Summary
Fixes the "Total Stock Buys" metric on the portfolio page which was always showing 0.

## Root Cause
The code was querying for  but PortfolioTransaction enum only has these types:
- , , , , 

There is no  type, so the query always returned 0.

## Fix
Changed to count completed buy orders directly:
```ruby
@portfolio.user.orders.completed.where(action: :buy).count
```

This correctly counts actual stock purchases that have been executed.

## Test Plan
- [x] Verified PortfolioTransaction enum types in model
- [x] Checked ExecuteOrder service to understand how orders create transactions
- [x] Tested in console with user who has completed orders
- [x] Confirmed count now shows correct number

Resolves #798

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/views/portfolios/show.html.erb
+++ b/app/views/portfolios/show.html.erb
@@ -24,10 +24,10 @@
           <div class="text-2xl font-semibold">$<%= sprintf('%.2f', @portfolio.cash_balance) %></div>
           <div><%# Trending up/down this month will go here %></div>
         </div>
-        <%# Total Stock Buys %>
+        <%# Total Stocks %>
         <div class="bg-white rounded-[22px] border border-black/30 py-4 px-5 flex flex-col justify-between min-h-[120px]">
-          <div class="text-sm text-gray-600">Total Stock Buys</div>
-          <div class="text-2xl font-semibold"><%= @portfolio.user.orders.completed.where(action: :buy).count %></div>
+          <div class="text-sm text-gray-600">Total Stocks</div>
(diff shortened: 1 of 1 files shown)


davidpgero (changes requested):
I think the `total stock buy` label here is misleading. 

The solution you provide is good, but I don't think we want that. 

We want the sum of all the shareholdings students have. 

image

Here, the count should be 5, not all the shares you even buy.

The PortfolioPosition already has shares that we can sum up here.

davidpgero:
@pollygee , the `Total stock buy` is misleading here. 

Can we rename it to `Total holding stocks` or something like that? 

And please, confirm my understanding. 

It should be the sum of all the holdings, not every stock purchase a student has ever made. 

@heyjosephme, I'm asking Polly to confirm my understanding here.

pollygee:
Thanks @davidpgero  yes you're correct total of the stocks owned in the portfolio.  I see how the name is misleading sorry about that.  I tried to make it clear in the ticket that it would be a total of the column from the table.  

We can change the name let's call it 
`Total Stocks`

Sorry for the miscommunication @heyjosephme are you ok to change this name and make the update?

davidpgero:
image

davidpgero (approved):
🚀

PR #813, opened 2025-10-27
Fix portfolio seed data to use transactions instead of current_position

### LINKED ISSUE #793: Update portfolio amount in seed data
We currently add $10,000 to a portfolio in the seed data but this doesn't make sense becuase it becomes the current position but we don't own any socks.  We need to add some transactions instead to make the cash value be an amount that would allow you to create buys and sells

So we should remove the current value that gets added and add a transaction or two that will actually create a balance that could be spend 

Image

heyjosephme:
## Summary
- Fixed portfolio seed data to use proper transactions instead of directly setting current_position
- Portfolio balances are now calculated from transactions (deposits, earnings, orders)
- Makes seed data realistic and matches how production portfolios work

## Changes
- Removed `current_position: 10_000.0` from Student and Mike user creation in `db/seeds/partials/users.rb`
- Added `PortfolioTransaction` deposit for Student user in `db/seeds/partials/portfolio_transactions.rb`
- Portfolio balances now correctly calculated from transaction history

## Root Cause
Previously, portfolios were created with `current_position: 10_000.0` directly set, which:
- Showed $10,000 balance without any transaction history
- Didn't reflect how real portfolios work (transactions create balance)
- Was confusing for developers learning the system

## Test Plan
- [x] Run `bin/dc rails db:reset` to test seed data
- [x] Verify Student user has transactions and calculated balance
- [x] Verify Mike user has transactions and calculated balance
- [x] Confirm portfolios work correctly in development

## Testing Results
```
Student cash: $9,932.03 (from 1 deposit transaction + pending orders)
Mike cash: $24,712.04 (from 10 transactions including deposits, withdrawals, and orders)
```

Resolves #793

### DIFF
--- a/db/seeds/partials/portfolio_transactions.rb
+++ b/db/seeds/partials/portfolio_transactions.rb
@@ -1,5 +1,26 @@
 # frozen_string_literal: true
 
+# Create transactions for Student user
+student = User.find_by(email: "student@example.com")
+
+if student
+  portfolio = Portfolio.find_or_create_by(user: student)
+
+  # Initial deposit to give student starting balance
+  PortfolioTransaction.create(
+    portfolio: portfolio,
+    transaction_type: :deposit,
+    amount_cents: 10_000_00
+  )
+
+  puts "Seeded portfolio transactions for Student user"
+  Rails.logger.info "Seeded portfolio transactions for Student user"
+else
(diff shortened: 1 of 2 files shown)


davidpgero (approved):
It looks good! Thank you! 🙇

Issue #797, opened 2025-10-22
Allow gradebook to be finalize even if we don't have grades for everyone


pollygee:
Currently you must enter grades for everyone before you can finalize a gradebook
there are times when we don't have access to grades for a student

Allow the grade book to be finalized even if we don't have data for everyone
Add a confirm message when we finalize to remind the admin that it can't be undone.

pollygee:
Yes!  Thanks

PR #830, opened 2025-11-01
Grade book finalizable regardless student grade entry

zulkiflizainuddin:
# Pull Request

## Summary
- Allow grade book to be finalized even if student grade entries are incomplete
- Add a confirm message when finalize to remind admin that it can't be undone.

## Related Issue
Resolves  (https://github.com/rubyforgood/stocks-in-the-future/issues/797)

## Changes
- Update logic on `GradeBooksController#finalize`
- Update styling class on  `grade_books/_finalize_button.html.erb`

## Checklist
- [x] CI checks passing

## Notes

### DIFF
--- a/test/controllers/grade_books_controller_test.rb
+++ b/test/controllers/grade_books_controller_test.rb
@@ -89,6 +89,22 @@ class GradeBooksControllerTest < ActionDispatch::IntegrationTest
     assert @grade_book.completed?
   end
 
+  test "finalize grade book with incomplete grade entry (empty string) and no previous quarter" do
+    sign_in(create(:admin))
+
+    @grade_book.grade_entries.first.update!(
+      math_grade: "", reading_grade: "", attendance_days: 30
+    )
+
+    post finalize_classroom_grade_book_path(@classroom, @grade_book)
+
+    assert_redirected_to classroom_grade_book_path(@classroom, @grade_book)
+    assert_equal "Grade book finalized. Funds have been distributed.", flash[:notice]
+
+    @grade_book.reload
+    assert @grade_book.completed?
+  end
+
   test "finalize grade book with complete grade entry and no previous quarter" do
     sign_in(create(:admin))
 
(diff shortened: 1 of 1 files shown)


test/controllers/grade_books_controller_test.rb
```diff
@@ -71,7 +71,7 @@ def setup
     assert @grade_book.verified?
   end
 
-  test "finalize does not finalize if entries are incomplete" do
+  test "finalize does finalize even if entries are incomplete" do
```
davidpgero: Maybe `finalize incomplete grade entries` ?

config/locales/en.yml
```diff
@@ -31,7 +31,7 @@ en:
         portfolio: Portfolio
   grade_books:
     finalize:
-      incomplete: 'Cannot finalize: Some entries are incomplete.'
+      incomplete: 'Cannot finalize: Grade book already finalized.'
```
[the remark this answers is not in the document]
zulkiflizainuddin: I've addressed in commit https://github.com/rubyforgood/stocks-in-the-future/pull/830/commits/a058cddcd811b01e44f3a012159efea7a25f146b

zulkiflizainuddin:
> Thank you for the improvement! 🙇
> 
> All in all, the solution is great! I like that!
> 
> It's just a bug we need to fix before we merge it.
> 
> When I tested, I ran into a bug.
> 
> Since we removed `finalizable?`, nil grades can now reach the `improved_grade?` method, which causes a crash:
> 
> `undefined method '<' for nil`
> 
> What we can do is add a guard for `math_improvement_earnings` and `reading_improvement_earnings`.
> 
> Eg.
> 
> ```ruby
> def math_improvement_earnings(previous_entry)
>     return 0 unless previous_entry
>     return 0 if math_grade.nil? || previous_entry.math_grade.nil?
> 
>     improved_grade?(math_grade, previous_entry.math_grade) ? EARNINGS_FOR_IMPROVED_GRADE : 0
>   end
> ```
> 
> And you can apply the same pattern to `reading_improvement_earnings`.
> 
> And I saw that you removed the unused test, which is excellent, but there is no additional test that catches issues like this.
> 
> For tests, please add:
> 
> * Test case: finalize grade book with nil math/reading grades (no previous quarter)
> * Test case: finalize grade book with nil grades when the previous quarter had complete grades
> 
> Can you fix the issue and add test cases, please?
> 
> After the changes, it's ready to merge 🚀

Yeah, will do and thanks for the detailed suggestions!!

app/models/grade_entry.rb
```diff
@@ -36,12 +34,14 @@ def attendance_perfect_earnings
 
   def math_improvement_earnings(previous_entry)
     return 0 unless previous_entry
+    return 0 if math_grade.nil? || previous_entry.math_grade.nil?
```
davidpgero: Oh, it looks like I was wrong. 

When I tested, I ran into an issue when the grades were an empty string. 

After I save a grade like this: 
image

student's math and reading are empty strings instead of nil.

So it should be `#blank?`, not `nil?`.

davidpgero (changes requested):
Thank you for the update! 🚀 

When I tested, I ran into an empty string issue. 

Can you update to catch nil and empty strings? 

We can update the test cases as well to cover this new empty-string case.

PR #853, opened 2025-11-08
Feature/multiple classroom memberships

### LINKED ISSUE #750: Allow students to belong to multiple classrooms
Currently we only allow student to belong to one classroom.  In speaking with the stakeholders it's become clear that they need to keep the data of which classroom a student previously belonged to.

We are adding an archived button to classrooms to help us with the multiple classes a student will have been a part of.

For this ticket we just want to allow students to belong to multiple classrooms.
On the admin side, we should add a table on the student show page to show a classroom history table that shows the previous class name, teacher  and school year that they belonged to.

Barsu5489:
# Student multiple classrooms memberships

## Summary
Implement many-to-many relationship for students and classrooms using a `StudentClassroom` join table, mirroring the existing `TeacherClassroom` pattern. Students can now belong to multiple classrooms, with archived status tracking for classroom history.

## Related Issue
Fixes #750 

## Changes
- Create `StudentClassroom` join table with `archived` flag for tracking classroom history
- Add `StudentClassroom` model with proper associations and uniqueness validation
- Update `Student` model to use `has_many :classrooms, through: :student_classrooms`
- Update `Classroom` model to use `has_many :students, through: :student_classrooms`
- Migrate existing `classroom_id` data to new join table
- Add model validations to catch invalid classroom references early
- Update student factory to create join records automatically
- Add comprehensive test coverage for new associations and behavior

## Checklist
- [ ] Issue is assigned (commenting on the issue page is needed)
- [ ] Issue link added to the PR's description
- [ ] Branch created from main
- [ ] Commits are small and descriptive
- [ ] Ran linter and fixed issues
- [ ] Ran tests and all tests pass
- [ ] CI checks passing
- [ ] Review requested from team members

## Notes

### DIFF
--- a/.better-html.yml
+++ b/.better-html.yml
@@ -1,0 +1,3 @@
+---
+allow_single_quoted_attributes: true
+allow_unquoted_attributes: false
--- a/.erb_lint.yml
+++ b/.erb_lint.yml
@@ -1,0 +1,25 @@
+---
+EnableDefaultLinters: true
+linters:
+  ErbSafety:
+    enabled: true
+    better_html_config: .better-html.yml
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/InitialIndentation:
+        Enabled: false
+      Layout/LineLength:
+        Enabled: false
+      Layout/TrailingEmptyLines:
+        Enabled: false
+      Naming/FileName:
+        Enabled: false
+      Style/FrozenStringLiteralComment:
+        Enabled: false
+      Lint/UselessAssignment:
+        Enabled: false
+      Rails/OutputSafety:
+        Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ group :development, :test do
   gem "brakeman", require: false
   gem "bundler-audit", require: false
   gem "debug", platforms: %i[mri windows]
+  gem "erb_lint", require: false
   gem "pry", "~> 0.15.2"
   gem "pry-byebug"
   gem "rubocop"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,6 +85,13 @@ GEM
     ast (2.4.3)
     base64 (0.3.0)
     bcrypt (3.1.20)
+    better_html (2.2.0)
+      actionview (>= 7.0)
+      activesupport (>= 7.0)
+      ast (~> 2.0)
+      erubi (~> 1.4)
+      parser (>= 2.4)
+      smart_properties
     bigdecimal (3.3.1)
     bindex (0.8.1)
     bootsnap (1.19.0)
@@ -129,6 +136,13 @@ GEM
     docile (1.4.1)
     drb (2.2.3)
     erb (6.0.0)
+    erb_lint (0.9.0)
+      activesupport
+      better_html (>= 2.0.1)
+      parser (>= 2.7.1.4)
+      rainbow
+      rubocop (>= 1)
+      smart_properties
     erubi (1.13.1)
     et-orbi (1.4.0)
       tzinfo
@@ -375,6 +389,7 @@ GEM
     simplecov-html (0.13.1)
     simplecov_json_formatter (0.1.4)
     sin_lru_redux (2.5.2)
+    smart_properties (1.17.0)
     solid_queue (1.2.4)
       activejob (>= 7.1)
       activerecord (>= 7.1)
@@ -451,6 +466,7 @@ DEPENDENCIES
   debug
   devise (~> 4.9)
   discard (~> 1.4)
+  erb_lint
   factory_bot_rails
   faker (~> 3.5.2)
   font-awesome-rails
--- a/README.md
+++ b/README.md
@@ -63,6 +63,22 @@ Use the **username** and **password** to log in and test the application locally
 
 Access the app via `localhost:3000`
 
+## Linting
+
+This project uses [erb_lint](https://github.com/Shopify/erb-lint) to catch ERB template errors during development.
+
+To check all ERB templates:
+
+```console
+$ bin/dc bundle exec erb_lint --lint-all
+```
+
+To see autocorrectable issues:
+
+```console
+$ bin/dc bundle exec erb_lint --lint-all --autocorrect
+```
+
 ## Contributing
 
 To understand the project better, read the [project documentation](docs/README.md).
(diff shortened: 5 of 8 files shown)


app/models/student.rb
```diff
@@ -18,4 +20,10 @@ def set_default_email
   def ensure_portfolio
     create_portfolio!(current_position: 0) if portfolio.blank?
   end
+
+  def classroom_must_exist
```
davidpgero: Can you help me a bit with that? 

After this change, we don't have `classroom` anymore, only `classrooms` (plural).

db/migrate/20251108191257_migrate_classroom_id_to_student_classrooms.rb
```diff
@@ -0,0 +1,11 @@
+class MigrateClassroomIdToStudentClassrooms < ActiveRecord::Migration[8.0]
+  def up
+    User.where(type: "Student").where.not(classroom_id: nil).each do |student|
+      StudentClassroom.find_or_create_by(student_id: student.id, classroom_id: student.classroom_id)
+    end
+  end
+
+  def down
```
davidpgero: We cannot reverse it after we've added more than one classroom per student, but now we can. 

Anyhow, I agree not to add it. 

But there should be a 3rd migration which removes the classroom_id on the user, right?

test/factories/users.rb
```diff
@@ -16,6 +16,11 @@
     sequence(:username) { |n| "student_#{n}" }
     sequence(:email) { |n| "student_#{n}@example.com" }
 
+    after(:create) do |student, _evaluator|
+      if student.classroom && !student.student_classrooms.exists?(classroom: student.classroom)
```
davidpgero: after we remove the classroom on the user, it will break.

davidpgero (dismissed):
👋 @Barsu5489, thank you for your hard work on this one! 

I have a couple of comments. 

I don't see any seed, controller, or view changes. 

Are there places in the app that still reference `classroom` (singular) that need to be updated to `classrooms` (plural)?

db/schema.rb
```diff
@@ -10,39 +10,39 @@
...
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
 
   create_table "action_text_rich_texts", force: :cascade do |t|
-    t.string "name", null: false
     t.text "body"
-    t.string "record_type", null: false
-    t.bigint "record_id", null: false
     t.datetime "created_at", null: false
+    t.string "name", null: false
```
davidpgero: these change shouldn't be here 🤔

davidpgero:
Can I pick your brains quickly, @pollygee and @Barsu5489?

I'm thinking we can split this issue into two. 🤔 

- First ticket is the current one. It's adding a necessary model and connection. 
- Second ticket, follow-up, remove the classroom for the user (student + teacher)

My other thought is that we can call this join `Enrollment`. 

- Multiple students can enroll in a classroom.
- Multiple teachers can enroll in the classroom. 

If we have a `role` in the enrollment, we can easily determine whether it's a teacher-student connection.

pollygee:
I like Enrollment. I think it’s more intuitive and we have the same concept for teacher and student.

app/models/student.rb
Barsu5489: ImportStudentService still uses classroom_id directly.

We can refactor it in follow-up PR.

db/migrate/20251108191257_migrate_classroom_id_to_student_classrooms.rb
Barsu5489: Agreed. Will handle the classroom_id removal after we refactor ImportStudentService to use the join table.

Barsu5489:
I agree, enrollment makes sense. I think it deserves its own PR.

Once this is approved we can tackle enrollment as a follow-up.

davidpgero:
> I agree, enrollment makes sense.

@Barsu5489 I don't follow it now 🤔 

Can you help me a bit? 

For me, this pr looks like a good candidate for enrollment. 
Then we can create a follow-up on removing the classroom_id. 

Do you mean we should go
1) with this StudentClassroom
2) Then refactor it to Enrollment
3) Then remove the classroom_id? 

I'm not convinced we need the StudentClassroom state for this process. 

It looks better to start with Enrollment instead and add the classroom to the enrollment, even if we don't use it anywhere. 

So we're in a better state when we can remove the old ones. 

Maybe I just miss something here.

Barsu5489:
> > I agree, enrollment makes sense.
> 
> @Barsu5489 I don't follow it now 🤔
> 
> Can you help me a bit?
> 
> For me, this pr looks like a good candidate for enrollment. Then we can create a follow-up on removing the classroom_id.
> 
> Do you mean we should go
> 
>     1. with this StudentClassroom
> 
>     2. Then refactor it to Enrollment
> 
>     3. Then remove the classroom_id?
> 
> 
> I'm not convinced we need the StudentClassroom state for this process.
> 
> It looks better to start with Enrollment instead and add the classroom to the enrollment, even if we don't use it anywhere.
> 
> So we're in a better state when we can remove the old ones.
> 
> Maybe I just miss something here.

@davidpgero 
 No you're not missing anything. The only reason I wanted it merged is because it was a completed, but I will rename it `Enrollment`.

test/models/student_test.rb
```diff
@@ -15,6 +15,34 @@ class StudentTest < ActiveSupport::TestCase
...
+    student.enrollments.create!(classroom: classroom2)
+
+    assert_equal initial_classroom_count + 2, student.classrooms.count
+    assert_includes student.classrooms, classroom1
+    assert_includes student.classrooms, classroom2
+  end
+
+  test "can have archived classroom in history" do
+    student = create(:student)
+    initial_count = student.enrollments.count # Factory creates 1 default join record
```
davidpgero: That's the reason I would avoid adding after create to the factory. 

It's much easier if you need to add it, and you don't need to remember it. 

What do you think?

Barsu5489:
@davidpgero and @pollygee. I am happy to take on the follow-up ticket right after this merges. I can start immediately.

One thing I need help with:
The CI/CD pipeline is showing test failures but all tests pass locally. This just appeared, everything was green and ready to merge earlier.

pollygee:
@Barsu5489 we chatted in slack and @dpowers1 is around so we can merge this once you have passing tests and he will keep an eye out for the follow up PR!  

Once you have passing tests you can tag @dpowers1 and I and one of us will approve and merge.  When you have a follow up PR ping @dpowers1 and he will check it out! 

Thanks so much.   Sorry for the delay several of us are out of pocket for a bit here.  Thanks for your help on the project and for your patience.

davidpgero:
follow-up ticket: https://github.com/rubyforgood/stocks-in-the-future/issues/870

Issue #847, opened 2025-11-06
Issue: Refactor PortfolioTransaction reason field to use enum instead of strings


davidpgero:
The `reason` field in `PortfolioTransaction` stores full string values in the database (e.g., "Earnings from Math", "Transaction Fees").

Converting to an enum will improve query, allow easy translation, and maintain consistency. 

## Current Implementation

   https://github.com/rubyforgood/stocks-in-the-future/blob/main/app/models/portfolio_transaction.rb#L9-L17

   https://github.com/rubyforgood/stocks-in-the-future/blob/main/app/models/portfolio_transaction.rb#L28-L30

   https://github.com/rubyforgood/stocks-in-the-future/blob/main/db/schema.rb#L142

## Desired Implementation

   **Model:**
   ```ruby
   enum :reason, {
     math_earnings: 0,
     reading_earnings: 1,
     attendance_earnings: 2,
     grade_earnings: 3,  # Deprecated, will be removed in future
     transaction_fees: 4,
     awards: 5,
     administrative_adjustments: 6
   }

   # Remove reason_humanized method (enum provides this via i18n)
   ```

   **Schema:**
   ```ruby
   t.integer "reason"
   ```

   **Translations (config/locales/en.yml):**
   ```yaml
   en:
     activerecord:
       attributes:
         portfolio_transaction:
           reasons:
             math_earnings: "Earnings from Math"
             reading_earnings: "Earnings from Reading"
             attendance_earnings: "Earnings from Attendance"
             grade_earnings: "Earnings from Grades"
             transaction_fees: "Transaction Fees"
             awards: "Award"
             administrative_adjustments: "Administrative Adjustment"
   ```

## Acceptance Criteria

   - [ ] `reason` column migrated from `string` to `integer`
   - [ ] Rails enum defined with numeric mappings (0-6)
   - [ ] i18n translations added for all reason values
   - [ ] `reason_humanized` method removed (use `human_enum_name(:reason)` or similar)
   - [ ] Data migration converts existing string values to enum integers
   - [ ] All existing reason values in database are successfully migrated

   ## Technical Considerations

   **Migration Strategy:**
   1. Create migration to add temporary `reason_enum` integer column
   2. Backfill data: map existing string values to enum integers
   3. Verify all records migrated successfully
   4. Drop old `reason` column, rename `reason_enum` to `reason`
   5. Update model with enum definition

   **Data Migration Mapping:**
   ```ruby
   # In migration
[...]

PR #866, opened 2025-11-11
Refactor PortfolioTransaction reason to use enum

JohnReyGimenez:
## Summary
This PR refactors the `PortfolioTransaction.reason` field from a `string` to an `integer` enum.

## Related Issue
Resolves https://github.com/rubyforgood/stocks-in-the-future/issues/847

## Changes
- Created migration to convert `portfolio_transactions.reason` from `string` to `integer`
- Updated `PortfolioTransaction` model to use `enum` and remove old `REASONS` constant
- Added i18n translations for the new `reason` enum values in `en.yml`
- Updated `DistributeEarnings` service to use the new enum keys
- Updated `TransactionFeeProcessor` service to use the correct `:transaction_fees` enum key
- Updated `Admin::Students` form dropdown to populate from the new `enum`
- Updated `DistributeEarningsTest` to check against new enum string values
- Removed obsolete test for the deleted `REASONS` constant from `PortfolioTransactionTest`

## Checklist
- [X] Issue is assigned 
- [X] Issue link added to the PR's description
- [X] Branch created from main
- [X] Commits are small and descriptive 
- [X] Ran linter and fixed issues
- [X] Ran tests and all tests pass
- [X] CI checks passing
- [ ] Review requested from team members

## Notes
Just a heads up btw, i noticed that `TransactionFeeProcessor` was using an incorrect key (`:transaction_fee` instead of `:transaction_fees`) before my changes so yeah.. just pointing that out


app/services/transaction_fee_processor.rb
```diff
@@ -23,8 +23,10 @@ def execute
   def add_transaction_fee(order)
     return if @users.include?(order.portfolio.user_id)
 
-    order.portfolio.portfolio_transactions.fees.create!(amount_cents: PortfolioTransaction::TRANSACTION_FEE_CENTS,
-                                                        reason: PortfolioTransaction::REASONS[:transaction_fee])
+    order.portfolio.portfolio_transactions.fees.create!(
+      amount_cents: PortfolioTransaction::TRANSACTION_FEE_CENTS,
+      reason: :transaction_fees
```
davidpgero: I think it was incorrect in the original form 🤔 

We don't have `transaction_fees` reason. 

Fee is a transaction type

test/services/distribute_earnings_test.rb
```diff
@@ -64,17 +64,17 @@ class DistributeEarningsTest < ActiveSupport::TestCase
     expected_math = GradeEntry::EARNINGS_FOR_B_GRADE + GradeEntry::EARNINGS_FOR_IMPROVED_GRADE
     expected_reading = GradeEntry::EARNINGS_FOR_A_GRADE + GradeEntry::EARNINGS_FOR_IMPROVED_GRADE
 
-    attendance_transaction = transactions.find { |t| t.reason == PortfolioTransaction::REASONS[:attendance_earnings] }
+    attendance_transaction = transactions.find { |t| t.reason == "attendance_earnings" }
```
davidpgero: `attendance_transaction = transactions.find { |t| t.attendance_earnings? }`

JohnReyGimenez:
Yeah i completely missed the migration file 😭 ill work on your suggestions when im available. Thank you!

davidpgero (changes requested):
It looks good! I like it!

There is one issue with the translation key: `reason` vs `reasons`. 

The second one is that the migration assumes that every portfolio transaction has a reason, which is not true. 

Can you take a look at it, please?

db/migrate/20251108133510_convert_portfolio_transaction_reason_to_enum.rb
```diff
@@ -0,0 +1,55 @@
+class ConvertPortfolioTransactionReasonToEnum < ActiveRecord::Migration[8.1]
```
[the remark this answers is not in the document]
JohnReyGimenez: Ah, i was confused for a bit about what you meant but i see it now. You were right, it would have broken deposit and fee. I've added allow_nil: true to the reason enum to allow nil values. Thanks for spotting that!

JohnReyGimenez:
Regarding the migration, it should be safe because I added this validation check that was specifically designed to handle this: if PortfolioTransaction.where.not(reason: nil).where(reason_enum: nil).exists?

The .where.not(reason: nil) part filters out all transactions where the reason is nil (like deposit and fee).

I was worried about this too, so I tested it specifically to make sure.

First, I rolled back and manually set up the before state to simulate production. I created 2 records with string reasons (the ones to migrate) and left 9 records with nil reasons (the ones to ignore).

This was the state of the database before running the migration:

image

Then, I ran the migration and checked the database again:

image

I hope this clears it up. If there are any more problems or things I missed, please let me know. Thanks again for all the feedback!

dpowers1 (approved):
Looks good!

Issue #782, opened 2025-10-21
Allow classroom to have multiple grade levels


pollygee:
We recently learned that there are after school classes and lunch clubs.  We can cover these by calling them a class.  But the biggest difference is they will have multiple grades in the classrooms.

So we need to allow multiple grades in a classroom and update the admin pages new/show/index to show the multiple grades or add multiple grades.  They should behave like we have for teachers and students.

PR #901, opened 2025-12-05
feat: add Grade model and ClassroomGrade join table with admin dashbo…

gr4yt3x:
## Summary
This PR is the first part of a two-PR implementation to support classrooms having multiple grade levels.
Following @davidpgero’s recommendation, this PR focuses only on creating the underlying structure:

- Adds the Grade model
- Adds the ClassroomGrade join model
- Creates the necessary migrations
- Adds model associations and validations
- Adds Administrate dashboard and controller
- Adds factories and full test coverage

A second PR will follow to update the application logic, UI, and references to use the new associations.

## Related Issue
Part of #782

## Checklist
- [x] CI checks passing

### DIFF
--- a/app/dashboards/grade_dashboard.rb
+++ b/app/dashboards/grade_dashboard.rb
@@ -61,7 +61,7 @@ class GradeDashboard < Administrate::BaseDashboard
   # Overwrite this method to customize how grades are displayed
   # across all pages of the admin dashboard.
   #
-  # def display_resource(grade)
-  #   "Grade ##{grade.id}"
-  # end
+  def display_resource(grade)
+    "Grade ##{grade.id}"
+  end
 end
--- a/app/models/grade.rb
+++ b/app/models/grade.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class Grade < ApplicationRecord
-  has_many :classroom_grades, dependent: :destroy
+  has_many :classroom_grades, dependent: :restrict_with_error
   has_many :classrooms, through: :classroom_grades
 
-  validates :name, presence: true
-  validates :level, presence: true
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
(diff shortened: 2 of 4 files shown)


app/models/grade.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Grade < ApplicationRecord
+  has_many :classroom_grades, dependent: :destroy
+  has_many :classrooms, through: :classroom_grades
+
+  validates :name, presence: true
```
davidpgero: We have DB validation, which is excellent! 

But it does not give a helpful error message to the users.

Adding validation to the `name` and `level` is useful here. it can be in the follow-up ticket.

app/dashboards/grade_dashboard.rb
```diff
@@ -0,0 +1,67 @@
...
+  #   COLLECTION_FILTERS = {
+  #     open: ->(resources) { resources.where(open: true) }
+  #   }.freeze
+  COLLECTION_FILTERS = {}.freeze
+
+  # Overwrite this method to customize how grades are displayed
+  # across all pages of the admin dashboard.
+  #
+  # def display_resource(grade)
+  #   "Grade ##{grade.id}"
```
davidpgero: not a blocker:
I like to use the display resource, something useful

app/models/grade.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Grade < ApplicationRecord
+  has_many :classroom_grades, dependent: :destroy
```
davidpgero: I'm not sure we want to enable destroying 🤔 

Like teacher deletes 5th grade, it's a big issue, because we lose data.

```
has_many :classroom_grades, dependent: :restrict_with_error
```

davidpgero:
@gr4yt3x , I created a second-phase ticket. Do you want to take it? If yes, please comment on that so I can assign it.

Issue #869, opened 2025-11-16
More robust system of storing stock prices


pollygee:
Yesterdays stock price is still getting saved sometimes when it shouldn't.  If there aren't new stock prices there's been no trading, like on a holiday or saturday then we will show 0% change for all stocks.  We updated it to skip the rest of the job if it's a weekend day.  But we are still having some problems because technically we're fetching the prices after 1am, and this doesn't account for holidays either.  But on a saturday at 1 am we do want to fetch fridays prices.

Lets create a more robust system that checks to see if we have updated stock prices and bail if the stock market hasn't traded since we last did a pull.  And lets move to fetch the prices late at night rather than past midnight because that adds complexity.

PR #912, opened 2025-12-13
Add robust trading day tracking to stock price updates

heyjosephme:
## Summary
Implements a robust system to prevent storing duplicate stock prices when the market hasn't traded (holidays, weekends). The job now tracks the trading day from the API and only updates prices when the market has actually traded since the last update.

## Related Issue
Resolves https://github.com/rubyforgood/stocks-in-the-future/issues/869

## Changes
- Add `last_trading_day` date field to stocks table
- Extract trading day from Alpha Vantage API response (field `07. latest trading day`)
- Compare dates before updating: skip if market hasn't traded since last update
- Add comprehensive tests for trading day comparison logic
- Automatically handles weekends, holidays, and any market closures

## Technical Details
The Alpha Vantage API already returns the latest trading day in the response. We now:
1. Extract this field along with the price
2. Compare it to the stock's `last_trading_day`
3. Only update if the trading day is newer (or null for first run)
4. Save the trading day alongside the price

This eliminates the complexity around weekends/holidays/midnight timing mentioned in the issue.

## Note on Schedule Change
The issue mentioned moving from 1 AM to 11 PM to reduce complexity. I haven't included this change yet because:
- The trading day check makes timing less critical
- Keeps this PR focused on the core fix
- Can add schedule change if maintainers prefer

Happy to add the schedule change in this PR or a follow-up.

## Checklist
- [x] Issue is assigned
- [x] CI checks passing

## Notes
- This is a draft PR to get feedback on the approach before finalizing
- All tests pass locally
- Class length slightly over limit (101/100) - can extract to service if preferred

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/jobs/stock_prices_update_job_test.rb
+++ b/test/jobs/stock_prices_update_job_test.rb
@@ -7,6 +7,7 @@ class StockPricesUpdateJobTest < ActiveJob::TestCase
 
   setup do
     ENV["ALPHA_VANTAGE_API_KEY"] = "test_api_key"
+    @last_trading_date = Date.parse("2024-05-31")
 
     stub_request(:get, STOCK_URL_MATCHER)
       .with(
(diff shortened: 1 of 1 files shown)


davidpgero (commented):
I like this solution! I didn't know we have the last trading day information in the response 👍

PR #1038, opened 2026-02-19
Add Validation to Order Creation

### LINKED ISSUE #1010: Add insufficient funds/shares validation to ExecuteOrder service
## Summary

The `ExecuteOrder` service is missing validation for:
- Canceling buy orders when user has insufficient funds
- Canceling sell orders when user has insufficient shares

## What's Needed

1. Add validation logic to `ExecuteOrder` service to check:
   - For buy orders: user has enough cash balance to cover the purchase
   - For sell orders: user owns enough shares to sell

2. Cancel orders that fail validation instead of executing them

3. Add tests in `test/services/execute_order_test.rb` for:
   - Buy order with insufficient funds → order should be canceled
   - Sell order with insufficient shares → order should be canceled

Josiassejod1:
# Pull Request

## Summary
There isn't current validation on `ExecuteOrder` when there are insufficient funds or shares, thus, this PR adds some tests and checks.

## Related Issue
Resolves #1010 

## Changes
- Add a valid_order check in the ExecuteOrder service that leverages similar logic that already exists in the order model

## Checklist
- [X ] Issue is assigned (commenting on the issue page is needed)
- [X ] Issue link added to the PR's description
- [X ] Branch created from main
- [ X] Commits are small and descriptive
- [X ] Ran linter and fixed issues
- [ X] Ran tests and all tests pass
- [X ] CI checks passing

## Notes


app/services/execute_order.rb
```diff
@@ -54,4 +59,18 @@ def update_order_status
...
   end
+
+  def valid_order?
+    if order.buy?
+      order.sufficient_funds?
+    elsif order.sell?
+      order.sufficient_shares?
+    end
+
+    true
```
pollygee: because this is outside of the if block this function will always return true and that isn't what we want we'd want it to return the result of `sufficient_funds` or `sufficient_shares`.

test/services/execute_order_test.rb
```diff
@@ -146,4 +146,32 @@ class ExecuteOrderTest < ActiveSupport::TestCase
...
+    stock = create(:stock, price_cents: 100)
+    create(:portfolio_stock, portfolio: portfolio, stock: stock, shares: 5, purchase_price: 100)
+
+    order = build(:order, :pending, :sell, shares: 10, stock: stock, user: portfolio.user)
+
+    assert_raises(ActiveRecord::RecordInvalid) do
+      order.save!
+    end
+
+    assert order.errors[:base].first.include?("Cannot sell more shares than you own")
```
pollygee: Is this the error message we'd get I don't see it added as an error when we do the check.

app/services/execute_order.rb
Josiassejod1: good callout I need to call return

pollygee:
@AbuMareBear do you have any availability any time soon to take a look at these updates to the PR?

PR #1040, opened 2026-02-22
Address Test Suite Issues

### LINKED ISSUE #1039: Minor Test Suite Issues
### What's the issue?

- Some tests can be flaky due to specifying specific values for factories that use sequences. Note that this flakiness can be difficult to see as it is dependent on the number of parallel workers you run the suite with and only happens when running the full suite. See example:

Image
 
- 2 tests currently have no assertions and clutter the test output with warnings. They should be skipped to avoid this clutter 

Image

- One test has an incorrect assertion regarding the memorable password generation. The regex does not take into consideration the possibility of superhero names that include special characters:

Image

### Additional details

None of these issues are always present and are dependent on the test seed and also other factors.

sean-dickinson:
# Pull Request

## Summary
- Fixes some flaky tests in the test suite
- Updates the memorable password generator for an edge case

## Related Issue
Resolves #1039

## Changes
- Fixes a flaky test that was unnecessarily specifying grade levels when creating grades
- Skips the 2 tests that currently have no assertions
- Updates the MemorablePasswordGenerator to also remove the `'` character

## Checklist
- [x] CI checks passing

### DIFF
--- a/app/services/memorable_password_generator.rb
+++ b/app/services/memorable_password_generator.rb
@@ -2,9 +2,20 @@
 
 # TODO: more robust solution later.
 class MemorablePasswordGenerator
-  NUMBERS = (1..99).to_a.freeze
+  RANGE = (1..99)
 
-  def self.generate
-    (Faker::Superhero.name + rand(1..99).to_s + Faker::Superhero.name).delete(" ").delete("-").delete("'")
+  class << self
+    def generate
+      password = Faker::Superhero.name + rand(RANGE).to_s + Faker::Superhero.name
+      normalize password
+    end
+
+    private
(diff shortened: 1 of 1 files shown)


app/services/memorable_password_generator.rb
```diff
@@ -2,9 +2,20 @@
 
 # TODO: more robust solution later.
 class MemorablePasswordGenerator
-  NUMBERS = (1..99).to_a.freeze
+  RANGE = (1..99)
```
sean-dickinson: This constant wasn't being used, so I updated it to be something that actually could be used in the formula

Issue #477, opened 2025-09-05
Add sorting and filtering to the table views


pollygee:
we have several tables that would be helpful to be able to sort and filter by the column headers

pollygee:
It might be too early for this we don't know yet what the users would need for filters and sorting, I'm just guessing that we will need this at some point.

PR #1103, opened 2026-04-01
basic sorting on classroom and orders index tables

jewls618:
# Pull Request

## Summary
Adds basic table sorting to the Transactions and Classrooms tables for users. 

## Related Issue
Resolves https://github.com/rubyforgood/stocks-in-the-future/issues/477

## Changes
- Add an `apply_sorting` to the `ApplicationController`
- User `apply_sorting` in the `ClassroomsController` and `OrdersController`
- Add a `sort_link` to the `ApplicationHelper`
- User the `sort_link` in the related view templates

## Demo
https://www.loom.com/share/f1aeadd5ed8e4898886740222e653344

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -62,9 +62,11 @@ class Order < ApplicationRecord
     reorder(shares: safe_direction(direction))
   }
   scope :order_by_total_cost, lambda { |direction = :asc|
-    dir = safe_direction(direction).to_s.upcase
+    dir = safe_direction(direction)
+
+    expression = Arel.sql("stocks.price_cents * orders.shares")
     joins(:stock)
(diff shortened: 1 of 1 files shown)


app/controllers/classrooms_controller.rb
```diff
@@ -123,4 +124,20 @@ def check_classroom_eligibility
 
     redirect_to root_path, alert: t("application.access_denied.no_access")
   end
+
+  def apply_classroom_sorting(collection)
```
dpowers1: I think this could be a scope on the model. Let me know what you think!

jewls618: Moved to the model!

Issue #627, opened 2025-09-23
we're using username for teacher we really want to usse the email for username


pollygee:

sparsh9:
Hey @pollygee , I would like to take on this issue, can you explain what is needed to be done exactly?

PR #1104, opened 2026-04-01
Teacher email is now their username versus a unique username

ainsleyaye:
# Pull Request

## Summary
Teacher usernames are now associated as their email addresses. There is no longer a prompt for a username when creating a teacher. Display names for the teacher will also show the teachers full name versus the username.

## Related Issue
Resolves https://github.com/rubyforgood/stocks-in-the-future/issues/627

## Changes
- Teacher usernames are now automatically derived from their email address. When a teacher record is created or updated, the` username `field is set to match the email — admins no longer need to enter a separate username for teachers
- The teacher creation/edit form has been simplified to remove the username field; the email field hint now communicates that it doubles as the login credential
- Teacher display names now show the teacher's full name rather than their username/email
- A data migration back-fills `username = email `for any existing teacher records
- Tests have been updated throughout to reflect that teacher usernames are email addresses, and invalid-params scenarios now test against a blank email rather than a blank username

### DIFF
--- a/app/controllers/admin/teachers_controller.rb
+++ b/app/controllers/admin/teachers_controller.rb
@@ -98,7 +98,7 @@ module Admin
     end
 
     def teacher_params
-      params.expect(teacher: [:email, :name, :username, { classroom_ids: [] }])
+      params.expect(teacher: [:email, :name, { classroom_ids: [] }])
     end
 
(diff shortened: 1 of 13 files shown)


dpowers1 (approved):
Seems right to me!

Issue #1089, opened 2026-03-30
BUG: creating a classroom is not creating gradebooks


pollygee:
When you create a classroom it should create quarters the school year. 
We have bug that the quarters aren't getting created and therefore gradebooks aren't getting created. 

We need to fix the bug and add tests to cover this behavior.

![Image](https://github.com/user-attachments/assets/516ce083-89bb-4466-ac29-818d1d2476b3)

pollygee:
Thanks @wmccone

davidpgero:
Hello @wmccone , I'm just checking in. 

Did you have a chance to look at it? 

If you don't have time to tackle it, we can re-assign it. It's up to you. 

Thanks!

wmccone:
@davidpgero Sorry, this should have been marked as done. My fault.

davidpgero:
Oh, don't worry about it. Thanks! 🙇

PR #1105, opened 2026-04-01
BUG: creating a classroom is not creating gradebooks #1089

wmccone:
# Pull Request
BUG: creating a classroom is not creating gradebooks #1089 

## Summary
Currently the process to create gradebooks when a class room is created is broken. Quarters are not created when a classroom is created and gradebooks are not created as well.

## Related Issue
Resolves 
[1089](https://github.com/rubyforgood/stocks-in-the-future/issues/1089)

## Changes
- Add call to school year service object in the classrooms controller.

## Checklist
- [x] CI checks passing

## Notes

Per my earlier commits in this ticket, it could be worth it to remove the service object that creates quarters and just have quarters create automatically from the school year model. Since this is not my Repo I will not make the ultimate call here but it could be a good idea to create some rules around how we want to treat school years. There is a ton of data that may fall into school years, so I dont know if its ultimately worth it to allow users to fully delete these records.


pollygee (approved):
This looks good @wmccone thanks a bunch.  I think you might be right about removing the service object, this fixes the bug for now and we can think about what we might want to have long term in a followup ticket.  I think we also have a race condition here where potentially a different request could come in to create a classroom/ school_year while I am creating one which would result in a failure because we're both trying to create the school year but not info to the user.  We can handle that in a follow up too.

Thanks for fixing this bug!

PR #1117, opened 2026-04-12
Add insufficient funds/shares validation to ExecuteOrder service

### LINKED ISSUE #1010: Add insufficient funds/shares validation to ExecuteOrder service

heyjosephme:
# Pull Request

## Summary
Add execution-time validation to the `ExecuteOrder` service so that buy orders with insufficient funds and sell orders with insufficient shares are canceled instead of blindly executed.

## Related Issue
Resolves https://github.com/rubyforgood/stocks-in-the-future/issues/1010

## Changes
- Added `insufficient_funds?` check before executing buy orders — cancels if portfolio cash balance is negative
- Added `insufficient_shares?` check before executing sell orders — cancels if portfolio doesn't hold enough shares
- Added test for buy order cancellation when funds disappear after order creation
- Added test for sell order cancellation when shares disappear after order creation

## Screenshots (if applicable)
N/A — backend service change only

## Checklist
- [x] CI checks passing

## Notes
This closes a gap in the order execution flow: validations only ran at order creation/update time, but portfolio state can change between creation and execution (e.g., withdrawals, other orders executing first). The `ExecuteOrder` service now validates before proceeding and cancels orders that would overdraw funds or exceed available shares.

### DIFF
--- a/app/services/execute_order.rb
+++ b/app/services/execute_order.rb
@@ -62,7 +62,21 @@ class ExecuteOrder
   end
 
   def insufficient_funds?
-    portfolio.cash_balance.negative?
+    available_balance_cents = settled_balance_cents - other_pending_buy_costs
+    available_balance_cents < purchase_cost
+  end
+
+  def settled_balance_cents
+    transactions = portfolio.portfolio_transactions
+    transactions.deposits.sum(:amount_cents) +
+      transactions.credits.sum(:amount_cents) -
+      transactions.debits.sum(:amount_cents) -
+      transactions.withdrawals.sum(:amount_cents) -
+      transactions.fees.sum(:amount_cents)
+  end
+
+  def other_pending_buy_costs
+    order.user.orders.pending.buy.where.not(id: order.id).sum(&:purchase_cost)
   end
 
   def insufficient_shares?
--- a/app/services/execute_order.rb
+++ b/app/services/execute_order.rb
@@ -62,21 +62,7 @@ class ExecuteOrder
   end
 
   def insufficient_funds?
-    available_balance_cents = settled_balance_cents - other_pending_buy_costs
-    available_balance_cents < purchase_cost
-  end
-
-  def settled_balance_cents
-    transactions = portfolio.portfolio_transactions
-    transactions.deposits.sum(:amount_cents) +
-      transactions.credits.sum(:amount_cents) -
-      transactions.debits.sum(:amount_cents) -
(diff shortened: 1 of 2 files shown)


app/services/execute_order.rb
```diff
@@ -51,6 +61,14 @@ def update_order_status
     order.update!(status: :completed, portfolio_stock:, portfolio_transaction:)
   end
 
+  def insufficient_funds?
```
davidpgero, line 64: I'm thinking about adding validation here. 🤔  

It looks like we already have the same validation on the Order model. `order.valid?` should return `false` if there are not enough funds or shares. 

Do we want to cancel the order only in these two cases only and have the executor try the next day again in the other not valid cases? 

What do you think @heyjosephme ?

heyjosephme: Thanks for the review @davidpgero 

I looked into `order.valid?` but I don't think it applies here.

I fully agree the two if-blocks are duplication and can collapse them into one.

davidpgero: Ah, you're right, they're already created in order, but if they were valid when they were created, it does not make them valid when we execute them in sequence 👍

test/services/execute_order_test.rb
```diff
@@ -126,6 +126,37 @@ class ExecuteOrderTest < ActiveSupport::TestCase
     assert_equal [3, 5], portfolio_stocks.pluck(:shares).sort
   end
 
+  test "buy order with insufficient funds is canceled" do
+    portfolio = create(:portfolio)
+    portfolio.portfolio_transactions.create!(amount_cents: 10_000, transaction_type: :deposit)
+    stock = create(:stock, price_cents: 5000)
+    order = create(:order, :pending, :buy, shares: 1, stock: stock, user: portfolio.user)
+
+    # Simulate funds disappearing after order creation
```
davidpgero, line 135: good comment!

davidpgero (approved):
works well! thank you!

